PR #1, opened 2020-05-06
Add a missing decorator for state().

xorrkaz:
Without this, is_active() fails since lab.state is a pointer to a method
and not the property itself.

### DIFF
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -795,6 +795,7 @@ class Lab:
             self.wait_until_lab_converged()
         logger.debug("started lab: %s", self._lab_id)
 
+    @property
     def state(self):
         """
         Returns the state of the lab.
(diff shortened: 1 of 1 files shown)


xorrkaz:
> The documentation and also the associated test do not use this as a property but rather as a callable. So, I'm not 100% sure whether I want to change this.
> 
> Looking at other methods, the same logic could be applied but isn't (`is_active`, `details`, `has_converged`...)
> 
> For me, having this as a callable is probably more consistent... and have the `is_active` method fixed.

That's fair.  I will point out that node.state is also a property, but changing lab.state now would break the API.  Instead, I fixed lab.is_active().

rschmied:
I'm going to merge this in here. The bundled 2.1 release will have this included as well

rschmied:
this has been addressed in the 2.1.0 code.

PR #4, opened 2020-07-03
Export lab configuration to YAML 

WojciechowskiPiotr:
### SUMMARY ###
A new method for exporting the lab configuration as YAML

### DESCRIPTION ###
In my application (https://github.com/WojciechowskiPiotr/CMLNetKit) I need to get lab configuration from the CML2 server via API. I extended the virl2-client library by a new method to achieve this, but it would be useful to add it to official library.

The method requires one argument which is ``str`` with lab identification, returns either the YAML object or raises HTTPError if the object is not found or there is something wrong with the connection

### DIFF
--- a/virl2_client/virl2_client.py
+++ b/virl2_client/virl2_client.py
@@ -28,7 +28,6 @@ from functools import lru_cache
 from pathlib import Path
 from urllib.parse import urljoin, urlsplit, urlunsplit
 
-
 import pkg_resources
 import requests
 import urllib3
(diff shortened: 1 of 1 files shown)


rschmied:
You could use

```python
# cl being a client library instance
lab = cl.join_existing_lab("0991e2")
print(yaml.safe_load(lab.download()))
```

Which is basically equivalent.... The `lab` class has a `download` method.

WojciechowskiPiotr:
Ok, somehow missed this

PR #6, opened 2020-11-06
Add client/server version comparison dunder methods

ChristopherJHart:
This PR adds the following dunder methods to the virl2_client.Version class that allows for comparison of major, minor, and patch versions between two Version objects:

* `__gt__`
* `__ge__`
* `__lt__`
* `__le__`

Furthermore, the existing `__eq__` dunder method is slightly modified to check the compared object's class prior to performing any comparison of major, minor, or patch versions.

All unit tests are passing. Further unit tests have been added for each new method covering common use cases and expected failure scenarios.


rschmied:
thanks @ChristopherJHart for your contribution... As a team, we need to identify the proper way to keep things in sync as there's an internal repo where PRs are being committed to. And that is currently our "single source of truth".  I'll keep you posted.

rschmied:
I'm going to merge this... thanks @ChristopherJHart !

Issue #8, opened 2021-03-08
add_tag() in node.py results in multiple copies of the same tag being added


nsthompson:
When using node.py add_tag() because it is reading and subsequently appending the new tag to self._tags and then pushing the list to the API it results in multiple entries of the same tag being pushed to CML.

My test code is:

```python
        # Apply Tags to Nodes
        if len(node_tags) > 0:
            print(f'Tag Count for node {node_label} is {len(node_tags)}')
            print(f'Node Tags for node {node_label} are {new_node.tags()}')
            for tag in node_tags:
                print(f'Adding tag {tag} to node {node_label}...')
                print(f'Node Tags for node {node_label} are {new_node.tags()}')
                new_node.add_tag(tag)
                sleep(2)
                print(f'Node Tags for node {node_label} are {new_node.tags()}')
```

The output from a test where I'm attempting to add 3 tags (oob, routers, test) to the device shows as follows:

```shell
Tag Count for node dc_router01 is 3
Node Tags for node dc_router01 are []
Adding tag oob to node dc_router01...
Node Tags for node dc_router01 are []
Node Tags for node dc_router01 are ['oob']
Adding tag routers to node dc_router01...
Node Tags for node dc_router01 are ['oob']
Node Tags for node dc_router01 are ['oob', 'oob', 'routers']
Adding tag test to node dc_router01...
Node Tags for node dc_router01 are ['oob', 'oob', 'routers']
Node Tags for node dc_router01 are ['oob', 'oob', 'routers', 'oob', 'oob', 'routers', 'test']
```

I've fixed this with a modification to node.py to only push the single tag to the API endpoint as follows:

```python
    def add_tag(self, tag):
        current = self._tags
        if tag not in current:
            current.append(tag)
            # Push single tag to API
            self._set_node_property("tags", [tag])
```

And the output after the change:

```shell
Tag Count for node dc_router01 is 3
Node Tags for node dc_router01 are []
Adding tag oob to node dc_router01...
Node Tags for node dc_router01 are []
Node Tags for node dc_router01 are ['oob']
Adding tag routers to node dc_router01...
Node Tags for node dc_router01 are ['oob']
Node Tags for node dc_router01 are ['oob', 'routers']
Adding tag test to node dc_router01...
Node Tags for node dc_router01 are ['oob', 'routers']
Node Tags for node dc_router01 are ['oob', 'routers', 'test']
```

nsthompson:
[...]

PR #9, opened 2021-03-08
updated node.py to only add single tags when add_tag() is called

nsthompson:
Fix for issue #8.

Modified add_tag() to only push a single tag at a time to the API.


rschmied:
Thanks for your contribution, @nsthompson !

Issue #10, opened 2021-05-17
requests_toolbelt is required


xorrkaz:
virl2-client requires requests_toolbelt, but it is not listed in the poetry.lock or in the wheel/dist on PyPi.

The workaround is to manually install requests_toolbelt.

rschmied:
confirmed.  Need to figure out how to include the change

PR #11, opened 2021-05-18
Missing changes from Bitbucket

rschmied:
- added requests-toolbelt (missing dependency),
  resolves https://github.com/CiscoDevNet/virl2-client/issues/10
- also added pytest configuration files
- added fixtures for tests to pass
- added tests which were missing from Bitbucket
- applied black formatting
- GitHub test automation
- assure 3.5 compatibility
- remove hashes from requirements.txt


rschmied (commented):
need to add @xorrkaz as a reviewer

xorrkaz:
Looks good to me.  Could we also add 3.9 to the supported Python list?

xorrkaz (approved):
Thanks for the Python add.

xorrkaz:
Why pull out Python 3.9 now?

rschmied:
I've merged this now -- I'm still not sure why the "action workflows" fail even though they do pass locally. Maybe it's a resource issue (e.g. running out of minutes or something).  I wish there would be better indication of why things get cancelled instead of simply cancelling all workflows with no obvious reason.

PR #18, opened 2021-11-04
Return lab status in ClientLibrary

labeveryday:
## What does this PR do?

When using the `virl2_client.virl2_client.ClientLibrary` you are able to return all labs in a list and the labs as a list of lab objects. The issue is that when you return all labs it would great to also have the status. 

## Recommendations for how to test this, or anything you are worried about?

I thought about implementing this in the `get_lab_list()` since it already returns a list. But it only returns a list of lab IDs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist/
 docs/build/*
 *.pyc
 *.egg-info
+venv
--- a/virl2_client/virl2_client.py
+++ b/virl2_client/virl2_client.py
@@ -708,3 +708,27 @@ class ClientLibrary:
         response = self.session.get(url)
         response.raise_for_status()
         return response.json()
+    
+    def get_all_labs_status(self, show_all: bool=False) -> list:
+        """
+        Retrieves a list of all defined labs and their status.
+
+        :param show_all: Whether to get only labs owned by the admin or all user labs
+        :type show_all: bool
+        :returns: A list of tuples
+        :rtype: list[tuple]
+        """
+        # TODO: integrate this further with local labs - check if already exist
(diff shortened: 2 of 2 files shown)


xorrkaz:
Thanks, @labeveryday !  Just curious why you want to push this into the library vs. having your application support grabbing status.  This is what I do in cmlutils:

```python
    lab_ids = client.get_lab_list(all_users)
    for id in lab_ids:
        labs.append(client.join_existing_lab(id))
    ...
    # print(labs[0].state())
```

This then gives me all the lab properties.  If I were going to push some of this back to the library, I might have the method return the whole lab object as I might want something other than title or state (and I do in cmlutils), and this would save me a call to `join_existing_lab()`.

labeveryday:
> ```python
> lab_ids = client.get_lab_list(all_users)
>     for id in lab_ids:
>         labs.append(client.join_existing_lab(id))
>     ...
>     # print(labs[0].state())
> ```

As I was writing code to implement this I found that I had I had to do more work to get the status of all labs. So I created a function that tells me the details like `id`, `lab title`, and then the `status` of the lab. From there I can now start or stop my lab based off of id. Rather than going through all of the iterations.


Screen Shot 2021-11-04 at 5 01 29 PM

rschmied:
Thanks for your interest and contribution here, @labeveryday. I do agree with @xorrkaz here...  this is code that should go into the application side of things, not into the library itself.  it really depends on your specific use case what attributes of a lab is returned as part of your implementation here.  and i think it's easy enough to have this implemented on the application side of things, not inside of the library.

PR #27, opened 2023-03-21
SIMPLE-5198 persistent link conditioning

daniel-valent:
small fixes to PCL link conditioning to go along with CML implementation of conditioning persistence

### DIFF
--- a/virl2_client/models/link.py
+++ b/virl2_client/models/link.py
@@ -22,8 +22,12 @@ from __future__ import annotations
 
 import logging
 import time
+import warnings
 from functools import total_ordering
 from typing import TYPE_CHECKING, Optional
+
+from ..utils import check_stale, locked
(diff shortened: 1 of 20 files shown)


virl2_client/models/link.py
```diff
@@ -276,18 +276,18 @@ def set_condition(
...
+        when no condition is applied, but that was never the case.)
 
         :return: the applied link condition
         """
         url = self.base_url + "/condition"
         condition = self._session.get(url).json()
-        if condition is None:
-            return None
-
         keys = ["bandwidth", "latency", "jitter", "loss"]
```
tmikuska, line 291: I don't think that we need to filter keys here either (we control the returned data on CML server side).

tmikuska (approved):

rschmied (approved):
lgtm

Issue #12, opened 2021-07-29
don't do initial commands in pyats connect


mgdransfield:
Issue when using cml client to access an sd-wan device via pyats. The initial commands fail as they are not allowed on sd-wan device. The testbed is generated on the fly so I can't see a way to set init_exec_commands=[], init_config_commands=[] on the connect line in virl2_client/models/cl_pyats.py  run_command.

xorrkaz:
Seems like being able to pass `init_exec_commands` and `init_config_commands` to the `run_command()` and `run_config_command()` methods might do the trick (where those are passed to `dev.connect()`).  Maybe something like:

```python
def run_command(self, node_label, command, init_exec_commands=None, init_config_commands=None):
...
    extra_params = {}
    if init_exec_commands:
        extra_params["init_exec_commands"] = init_exec_commands
    if init_config_commands:
        extra_params["init_config_commands"] = init_config_commands

    pyats_device.connect(log_stdout=False, **extra_params)
```

mgdransfield:
def run_command(self, node_label, command, init_exec_commands=[], init_config_commands=[]):
...

    pyats_device.connect(log_stdout=False, init_exec_commands=init_exec_commands,init_config_commands=init_config_commands)

works although changes default behaviour

xorrkaz:
Yes, this does change the default behavior, which is why I suggested my approach above.  This way, clients can pass different initial commands to the two cl_pyats methods without changing the default.

rschmied:
can we close this?

PR #29, opened 2023-03-23
Do not do initial commands in pyats

tmikuska:
resolves: #12

### DIFF
--- a/virl2_client/models/cl_pyats.py
+++ b/virl2_client/models/cl_pyats.py
@@ -24,6 +24,8 @@ import io
 from typing import TYPE_CHECKING, Any, Optional
 
 if TYPE_CHECKING:
+    from pyats.topology import Device
+
     from .lab import Lab
 
 
@@ -91,7 +93,7 @@ class ClPyats:
         data.devices.terminal_server.credentials.default.password = password
         self._testbed = data
 
-    def _prepare_device(self, node_label: str):
+    def _prepare_device(self, node_label: str) -> Device:
         self._check_pyats_installed()
         try:
             pyats_device = self._testbed.devices[node_label]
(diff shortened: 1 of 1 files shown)


virl2_client/models/cl_pyats.py
```diff
@@ -110,29 +104,53 @@ def run_command(self, node_label: str, command: str) -> str:
...
+        :returns:  a dictionary of optional parameters to be executed with a command
+        """
+        params = {}
+        if init_exec_commands:
+            params["init_exec_commands"] = init_exec_commands
+        if init_config_commands:
+            params["init_config_commands"] = init_config_commands
+        return params
+
+    def run_command(self, node_label: str, command: str, init_exec_commands: Optional[list] = None, init_config_commands: Optional[list] = None) -> str:
```
virlos, line 126: Perhaps fix the lines that are too long

rschmied (commented):
code-wise this lgtm. I wonder what will break if we do not send the initial commands down to the device. are we ensuring that e.g. paging and logging is turned off? but maybe I misunderstand what the goal here is.

tmikuska:
> code-wise this lgtm. I wonder what will break if we do not send the initial commands down to the device. are we ensuring that e.g. paging and logging is turned off? but maybe I misunderstand what the goal here is.
> 
> edit: read up on #12 -- who / what ensures that for devices where the initial commands are working / do useful things those commands are actually sent?

1. I believe that PyATS module itself handles that.
2. The code here doesn't change the default behavior.  It just allows a user to override it.
3. I cannot say whether #12 was caused by improper SD-WAN setup/config or there's actually missing/broken support in PyATS.   I just created a PR with the proposed workaround.  We can mark it as won't do instead.

PR #32, opened 2023-04-03
SIMPLE-5280 auth management refactor

daniel-valent:
Ready for merge


PatrikMosko:
Looks good for me for now. We will se once I run the tests on it.

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,392 @@
...
+        value = AuthMethod(value)
+        self._session.put(self._METHOD_URL, json={"method": value.value})
+        self._method = value
+
+    @property
+    def manager(self) -> AuthMethodManager:
+        """
+        The manager for the current auth method.
+        """
+        # TODO: should we allow users to also configure currently inactive auth methods?
```
PatrikMosko, line 88: API allows you to only configure currently set method, same apply for UI. I therefore think user should also be only allowed to update currently activated auth_manager.

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,392 @@
...
+        Update a setting for this auth method.
+
+        :param setting: the setting to update
+        :param value: the value to set the setting to
+        """
+        if setting == "method":
+            raise ValueError(
+                f"To set method, use 'AuthManagement.method = {value}' instead."
+            )
+        # TODO: currently, method is required, so we have to provide it;
```
PatrikMosko, line 151: We should think 2 steps ahead in this case. Since there is a discussion about updating API schema and create separate schema for each method, which means each auth method is going to have its own (required) properties, it doesn't make any sense to allow user to update method WITHOUT updating every other property, if that's what you are referring to.

If you are talking about completely switching auth_manager when you switch method, I guess that makes sense, but as we already discussed, I would like to switch update_settings to private method and only allow user to change method by method property, which directly fixes this issue of yours.

virl2_client/models/auth_management.py
tmikuska: AFAIK, you can configure all LDAP properties and still set the method to "local" in a single request (via API only).  This is imo a bad behavior and will be restricted eventually (when? no idea)

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,392 @@
...
+        :param kwargs: keywords of settings
+        """
+        settings = {}
+        if settings_dict:
+            settings.update(settings_dict)
+        settings.update(kwargs)
+        if not settings:
+            raise TypeError("No settings to update.")
+        if settings.get("method", None):
+            # TODO: see update_setting above
```
PatrikMosko, line 194: I would like to preserve consistency with the way how API handles this, ie implement it as a PUT method.

You should be able to insert method here, along its required/supported property values.

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+        """
+        # TODO: should we allow users to also configure currently inactive auth methods?
+        #  not sure how useful that might or might not be, but if they can then that
+        #  changes the answer to the question in AuthMethodManager.update_setting
+        self.sync_if_outdated()
+        return self._managers[self._method]
+
+
+class AuthMethodManager:
+    _METHOD = ""
```
tmikuska, line 96: shouldn't this be None or a value for the AuthMethod enum?

virl2_client/models/auth_management.py
PatrikMosko: maybe you can future-proof it and add a check like:
`if settings.get("method", None) == "local" and len(settings) > 1:
    raise ValueError("Local auth method doesn't define extra parameters.")`

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+        self._settings = self._get_current_settings()
+        self._last_sync_time = time.time()
+
+    @property
+    def _session(self) -> Client:
+        return self._auth_management._session
+
+    def _get_current_settings(self) -> dict:
+        settings = self._session.get(self._CONFIG_URL).json()
+        del settings["method"]
```
tmikuska, line 124: why do we delete method?  I don't think it's a big deal.

virl2_client/models/auth_management.py
PatrikMosko: Currently, in the API it is possible to "update" local auth config by adding other properties, however these properties are only relevant for LDAP auth method. It should be changed in the future in the schema and currently it works only because we have no PATCH endpoint for changing auth properties.

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+        """
+        self.sync_if_outdated()
+        return self._settings[setting]
+
+    def get_settings(self) -> dict:
+        """Returns a dictionary of the settings of this authentication method."""
+        self.sync_if_outdated()
+        return self._settings.copy()
+
+    def _update_setting(self, setting: str, value) -> None:
```
tmikuska, line 140: value: Any?

virl2_client/models/auth_management.py
daniel-valent: Redundancy; you are supposed to get method from AuthManagement, not a manager. Also, if we split it away from the settings later, that would remove it from here anyway, so I'm just doing it ahead of time so that users don't expect it to stay.

virl2_client/models/auth_management.py
daniel-valent: A string is more convenient (which matters more since this is private) and honestly it does not matter what value we set as default, it's going to be overwritten by subclass anyway

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+        :param settings_dict: a dictionary of settings
+        :param kwargs: keywords of settings
+        """
+        settings = {}
+        if settings_dict:
+            settings.update(settings_dict)
+        settings.update(kwargs)
+        if not settings:
+            raise TypeError("No settings to update.")
+        if settings.get("method", None):
```
tmikuska, line 193: None is the default value.

virl2_client/models/auth_management.py
PatrikMosko: Yeah, I realized in my next comment. As you pointed out it should not be the case in the future, so I don't think it will be an issue if we already implement it properly in the PCL.

virl2_client/models/auth_management.py
daniel-valent: Right, pop raises by default, not get, thanks.

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+            # TODO: see update_setting above
+            raise ValueError("Settings include method.")
+        settings["method"] = self._METHOD
+        self._session.put(self._CONFIG_URL, json=settings)
+        self._auth_management._method = AuthMethod(settings["method"])
+        # If we didn't get the setting from the API when syncing, it is write-only,
+        # so we don't want to store it; we only update existing keys
+        storable_settings = {
+            shared_key: settings[shared_key]
+            for shared_key in settings.keys() & self._settings.keys()
```
tmikuska, line 203: keys() method is redundant here

virl2_client/models/auth_management.py
tmikuska: agreed

virl2_client/models/auth_management.py
```diff
@@ -0,0 +1,391 @@
...
+        if setting in self._settings:
+            self._settings[setting] = value
+
+    def update_settings(self, settings_dict: Optional[dict] = None, **kwargs) -> None:
+        """
+        Update multiple auth settings at once.
+        More efficient in batches than setting manager properties.
+        If passed a dictionary, reads the dictionary for settings.
+        If passed multiple keyword arguments, each is taken as a setting.
+        Can mix and match; keywords take precedence.
```
tmikuska, line 174: I think that this is an overkill for a method that should be used once at max if ever :) but sure

virl2_client/models/auth_management.py
tmikuska: > if settings.get("method", None) == "local" and len(settings) > 1: raise ValueError("Local auth method doesn't define extra parameters.")

We can, but this should be fixed on server/API side anyway.

daniel-valent: I'd prefer to just leave checks like that to the API. Whatever the user provides, we'll send it, and if the API complains, we raise. Problem solved.

virl2_client/models/auth_management.py
PatrikMosko: It is not that unnecessary, I would still keep it. However I think we should be able to switch method with this one too, if we provide "method" as an argument.

daniel-valent: 1. The method is definitely useful. Might be a bit overengineered, but you can e.g. pass it a dictionary of settings but override one using a keyword.
2. Strong disagree on allowing 'method' as argument. This is a method of managers. If you use LDAPManager, you should not be able to use it to switch to 'local'; that is what we already agreed on, isn't is?

virl2_client/models/auth_management.py
PatrikMosko: As we already agreed on other things, this should be and probably will be handled by API once the schema is fixed, therefore we can take the lazy approach and "do nothing" here.

daniel-valent: Ok, if we leave the errors to the API, we still have two identically simple options:
```python
client.auth_management.method = "ldap"
ldap_manager = client.auth_management.manager
client.auth_management.method = "local"
ldap_manager.verify_tls = True  # 200 OK, method is changed to ldap
```
or
```python
client.auth_management.method = "ldap"
ldap_manager = client.auth_management.manager
client.auth_management.method = "local"
ldap_manager.verify_tls = True  # 400 Bad Request
```
Honestly, this doesn't matter much either way, so feel free to merge @PatrikMosko @tmikuska or if you want this changed, just drop a comment back.

PatrikMosko: I think it should pass currently, if you are changing ldap-related properties even if local is set. Currently we can consider it a "feature", later, once the schema is fixed, there wont be any extra properties for local so it should fail then. So I would say neither of your options, just patch the property silently (even if local is set) and do nothing, that is how our API works now.

PR #33, opened 2023-04-06
SIMPLE-5280 pcl auth methods missing or in wrong place - redone

daniel-valent:
Previous version was suboptimal, has been redone, so a new PR


virl2_client/models/auth_management.py
```diff
@@ -21,34 +21,37 @@
...
+        """
         self.auto_sync = auto_sync
         self.auto_sync_interval = auto_sync_interval
         self._last_sync_time = 0.0
         self._session = session
-        self._method = AuthMethod.LOCAL
-        self._managers: dict[AuthMethod, Optional[AuthMethodManager]] = {
-            AuthMethod("local"): None,
-            AuthMethod("ldap"): LDAPManager(self),
+        self._settings = {}
```
tmikuska, line 51: No settings are set by default, we need to call sync or set {"method": "local"} here.  Otherwise requesting an attribute will fail with KeyError.

Or we can do this:
~~~
- self._settings[setting]
+ self._settings.get(setting)
~~~

tmikuska: There are a few occurrences of `self._settings["method"]` too.

daniel-valent: I don't see anywhere where _settings are accessed without calling sync_if_outdated first, do you?

virl2_client/models/auth_management.py
```diff
@@ -217,14 +174,33 @@ def test_current_auth(
...
+    def __init__(self, auth_management: AuthManagement):
+        self._auth_management = auth_management
+
+    def _check_method(self):
+        """This makes sure that storing the manager in a variable won't allow the user
+        to accidentaly set settings of other methods using this."""
+        if self._auth_management._get_setting("method") != self.METHOD:
+            raise MethodNotActive(f"{self.METHOD} is not the currently active method.")
+
+    def _get_setting(self, setting):
```
tmikuska, line 193:
```suggestion
    def _get_setting(self, setting: str) -> Any:
```
this is possible too, but probably an overkill especially when considering that more types will be added in future 
```
    def _get_setting(self, setting: str) -> str | bool | ResourcePool
```

virl2_client/models/auth_management.py
```diff
@@ -217,14 +174,33 @@ def test_current_auth(
...
+        self._check_method()
+        return self._auth_management._get_setting(setting)
+
+    def _update_setting(self, setting, value):
```
tmikuska, line 197:
```suggestion
    def _update_setting(self, setting: str, value: Any) -> None:
```

virl2_client/models/auth_management.py
daniel-valent: Right, the issue was when sync is off... Submitted a fix which syncs if settings are empty even with sync off, we need the initial data either way

virl2_client/models/auth_management.py
virlos: Not following the discussion, but I don't think the auth method setting in the client should use the ResourcePool object itself, just its ID as a string. You never know if the object still exists.

daniel-valent: That, and also this wouldn't be really helpful. I guess it tells you that no settings ask for any other types, but you still don't know whether a setting needs string or bool or RP.

tmikuska: resource_pool setter allows both RP and RP.id as value (look at the end of this file)

PR #38, opened 2023-06-07
SIMPLE-5479 option for setting definition read only property

daniel-valent:
5377 needs to merge first

### DIFF
--- a/virl2_client/models/node_image_definitions.py
+++ b/virl2_client/models/node_image_definitions.py
@@ -82,6 +82,30 @@ class NodeImageDefinitions:
         url = "node_definitions/" + definition_id + "/image_definitions"
         return self._session.get(url).json()
 
+    def set_image_definition_read_only(
+        self, definition_id: str, read_only: bool
+    ) -> None:
+        """
+        Set the read-only attribute of the image definition with the given ID.
+
+        :param definition_id: The ID of the image definition.
+        :param read_only: The new value of the read-only attribute.
+        """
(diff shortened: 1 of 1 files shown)


tmikuska:
> 5377 needs to merge first

Can you separate this change, so we can merge it before 5377?  I assume that it will take time to review 5377 and it also may be too risky for 2.6.

daniel-valent:
Hmm, I'll basically need to downgrade the url stuff, but yeah can do

daniel-valent:
Done, ready for review

tmikuska (approved):
lgtm thanks

virl2_client/models/node_image_definitions.py
```diff
@@ -82,6 +82,30 @@ def image_definitions_for_node_definition(self, definition_id: str) -> list[dict
...
+        self, definition_id: str, read_only: bool
+    ) -> None:
+        """
+        Set the read-only attribute of the image definition with the given ID.
+
+        :param definition_id: The ID of the image definition.
+        :param read_only: The new value of the read-only attribute.
+        """
+        url = f"image_definitions/{definition_id}/read_only"
+        self._session.put(url, json=read_only)
```
tmikuska, line 95: can we change these to return a response?  I just found out that both endpoints return a full def object.
```suggestion
        return self._session.put(url, json=read_only).json()
```

daniel-valent: Done.

PR #39, opened 2023-06-07
SIMPLE-5377 PCL general improvements

daniel-valent:


virl2_client/event_handling.py
```diff
@@ -66,60 +101,112 @@ def _parse_lab(self, event: Event) -> None:
         elif event.subtype == "state":
             self._parse_lab_state(event)
         else:
-            _LOGGER.warning("Received an invalid lab event (%s)", event.subtype)
+            _LOGGER.warning(f"Received an invalid lab event ({event.subtype})")
 
     @abstractmethod
     def _parse_lab_created(self, event: Event) -> None:
```
virlos, line 107: shouldn't these then be called _handle_lab_created and such?

virl2_client/event_handling.py
```diff
@@ -66,60 +101,112 @@ def _parse_lab(self, event: Event) -> None:
...
+        """
         if event.subtype == "created":
             self._parse_element_created(event)
         elif event.subtype == "modified":
             self._parse_element_modified(event)
         elif event.subtype == "deleted":
             self._parse_element_deleted(event)
         else:
-            _LOGGER.warning("Received an invalid element event (%s)", event.subtype)
+            _LOGGER.warning(f"Received an invalid element event ({event.subtype})")
```
virlos, line 155: unknown subtype rather than invalid. There's also "state" for state_changed. And the server event_handling.py has these three in all-caps (I guess you lowercase them somewhere).

virl2_client/event_handling.py
```diff
@@ -141,23 +228,6 @@ def parse_event(self, event: Event) -> None:
                     event.element = event.lab.get_link_by_id(event.element_id)
             except ElementNotFound:
                 return
```
virlos, line 230: You're not logging these because it's common that an event arrives for a recently-deleted element?

virl2_client/event_handling.py
```diff
@@ -195,13 +265,15 @@ def _parse_element_created(self, event: Event) -> None:
                 event.data["interface_b"],
             )
         else:
-            _LOGGER.warning("Received an invalid element type (%s)", event.element_type)
+            _LOGGER.warning(f"Received an invalid element type ({event.element_type})")
```
virlos, line 268: There are also annotations. And connector mappings (which really only have one generic event, but those aren't "invalid".
You could just ignore them with a maybe debug/info log.

virl2_client/models/authentication.py
```diff
@@ -33,67 +33,98 @@
...
+        """
+        Return the authentication token. If the token has not been set, it is obtained
+        from the server by calling the API's 'authenticate' endpoint.
+        """
         if self._token is not None:
             return self._token
 
-        auth_url = "authenticate"
-        base_url = self.client_library.session.base_url
+        base_url = self.client_library._session.base_url
```
virlos, line 69: _under access will be also reported by linters, mind you. You could reduce them by having a _session property on this object as well

virlos (commented):
Only went through first part of the changes yet.

virl2_client/models/lab.py
```diff
@@ -614,34 +641,35 @@ def _remove_node_local(self, node: Node) -> None:
...
         If you have a link object, you can also simply do::
 
             link.remove()
 
-        :param link: the link object or ID
-        :param wait: Wait for convergence (if left at default,
-            the lab wait property takes precedence)
+        :param link: The link object or ID.
+        :param wait: A flag indicating whether to wait for convergence.
+            If left at the default value, the lab's wait property takes precedence.
```
virlos, line 672: property is used instead.

virl2_client/models/lab.py
```diff
@@ -666,18 +694,18 @@ def _remove_link_local(self, link: Link) -> None:
...
         If you have an interface object, you can also simply do::
 
             interface.remove()
 
-        :param iface: the interface object or ID
-        :param wait: Wait for convergence (if left at default,
-            the lab wait property takes precedence)
+        :param iface: The interface object or ID.
+        :param wait: A flag indicating whether to wait for convergence.
+            If left at the default value, the lab's wait property takes precedence.
```
virlos, line 708: used instead. - 'takes precedence' to what?
Also, we _could_ in a different ticket default to _UNCHANGED, and support None as not wanting to wait no matter what lab has set. But that has compatibility implications.

virl2_client/models/lab.py
```diff
@@ -917,146 +935,144 @@ def sync_states(self) -> None:
...
+        """
+        url = self._url_for("check_if_converged")
         converged = self._session.get(url).json()
         return converged
 
     @check_stale
-    def start(self, wait: Optional[bool] = None) -> None:
+    def start(self, wait: bool | None = None) -> None:
+        """
+        Starts all the nodes and links in the lab.
```
virlos, line 985: Technically, it starts all nodes - it starts only links connected to nodes which needed to be started as far a I remember. 
Tomas - should we consider that a bug or feature, from a "start lab" perspective?

virl2_client/models/lab.py
```diff
@@ -1079,25 +1094,25 @@ def build_configurations(self) -> None:
...
+        with_node_configurations: bool | None = None,
         exclude_configurations=False,
     ) -> None:
         """
-        Synchronize current lab applying the changes that
-        were done in UI or in another ClientLibrary session::
-
-            lab.sync()
+        Synchronize the current lab, applying changes made
+        in the UI or another ClientLibrary session.
```
virlos, line 1102: "by other API users" instead? It's not just those two.

virl2_client/models/node.py
```diff
@@ -474,123 +584,194 @@ def wait_until_converged(
...
     @check_stale
-    def console_logs(self, console_id: int, lines: Optional[int] = None) -> dict:
-        query = "?lines=%d" % lines if lines else ""
-        url = self._base_url + "/consoles/%d/log%s" % (console_id, query)
+    def console_logs(self, console_id: int, lines: int | None = None) -> dict:
+        """
+        Get the console logs of the node.
+
+        :param console_id: The ID of the console.
+        :param lines: The number of lines to retrieve.
```
virlos, line 643: Does None mean all available?

virl2_client/models/node_image_definitions.py
```diff
@@ -24,81 +24,95 @@
 import pathlib
 import time
 import warnings
-from typing import TYPE_CHECKING, BinaryIO, Callable, Optional
+from typing import BinaryIO, Callable, TYPE_CHECKING
```
virlos, line 27: Doesn't isort put constants first?

virl2_client/models/users.py
```diff
@@ -20,45 +20,61 @@
...
+        """
+        Generate the URL for a given API endpoint.
+
+        :param endpoint: The desired endpoint.
+        :param **kwargs: Keyword arguments used to format the URL.
+        :returns: The formatted URL.
+        """
+        return _url_from_template(endpoint, self._URL_TEMPLATES, kwargs)
 
     def users(self) -> list:
```
virlos, line 52: so list[str], or we didn't do this elsewhere either?

virl2_client/event_handling.py
daniel-valent: These are for event type "lab_element_event", while state changes are event type "state_change", subtype "state" (idk why).
Capitalization is useless, so it is discarded in Event initialization.

virl2_client/event_handling.py
daniel-valent: Yep. When a node is deleted, the node deletion event arrives first, which locally deletes everything related to that node, so the slew of interface/link deletion events afterwards are useless, for example.

virl2_client/models/authentication.py
daniel-valent: Yeah, this is kind of a problem.
The underscore should definitely be used in a situation like this. The end user should never need to access the session, so it should be hidden from them - and this is the tool we have for that. It signalizes that it is an internal thing, that should have no use to the end user and may be changed at any moment. I don't, however, see any reason for the underscore to restrict us in any way. If anything about it changes, it is me making the change. It also very much has a use internally. So, what benefit could we possibly get from restricting its use between classes in the same module?
IMO, the linters are wrong in this case, and I don't know if there is any way to shut them up.

examples/sample.py
```diff
@@ -45,55 +49,74 @@
 )
 
 
-# setup the connection and clean everything
-cl = ClientLibrary("http://localhost:8001", "cml2", "cml2cml2", allow_http=True)
-cl.is_system_ready(wait=True)
+# Set up the CML 2 connection
+server_url = "http://localhost:8001"
+username = "cml2"  # Default username if not changed in CML instance
+password = "cml2cml2"  # Default password if not changed in CML instance
```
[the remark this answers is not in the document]
virlos: The thing after the colon is what I suggest the comment should read.

virl2_client/models/authentication.py
daniel-valent: That is not pedantic, you are right, there was no reason for that to be public. Fixed, thanks.
We have many, many cases of using private functions and accessing private attributes of other classes/instances withing the same library, such as `self._client_library._remove_lab_local(event.lab)`, where there is no other solution.

tests/test_client_library.py
```diff
@@ -24,6 +24,7 @@
 import re
 import sys
 from pathlib import Path
+from typing import Optional
```
tmikuska, line 27: unused

virl2_client/event_handling.py
```diff
@@ -22,111 +42,196 @@
...
+
+    def _handle_element(self, event: Event) -> None:
+        """
+        Handle lab element events.
+
+        :param event: An Event object representing the lab element event.
+        """
+        if event.element_type in ("annotation", "connectormapping"):
+            # These are not used in this client library
+            _LOGGER.debug(f"Received an unused element type: {event.data}")
```
tmikuska, line 158: return?

virl2_client/event_handling.py
```diff
@@ -135,47 +240,37 @@ def parse_event(self, event: Event) -> None:
...
-                #         event.lab_id,
-                #     )
-                # )
-
-        super().parse_event(event)
-
-    def _parse_lab_created(self, event: Event) -> None:
+                if event.subtype == "deleted":
+                    # Element was likely already deleted in a cascading deletion
+                    # (e.g. node being deleting and all its links and interfaces being
```
tmikuska, line 250: being deleted

virl2_client/models/authentication.py
```diff
@@ -118,6 +155,17 @@ def response_raise(response: httpx.Response) -> None:
 
 
 def make_session(base_url: str, ssl_verify: bool = True) -> httpx.Client:
+    """
+    Creates an httpx Client object with the specified base URL
```
tmikuska, line 159: Based on other changes in this PR, this should say "Create" instead of "Creates".  (though, I'm not sure if 3rd person singular is not the correct one - opinions?)

tmikuska:
Done in https://github.com/CiscoDevNet/virl2-client/pull/59.

PR #41, opened 2023-06-21
SIMPLE-4394 minor improvements / SIMPLE-5561 fixed test with events

daniel-valent:


virl2_client/event_handling.py
```diff
@@ -177,6 +177,14 @@ def _parse_lab_state(self, event: Event) -> None:
 
     def _parse_element_created(self, event: Event) -> None:
         new_element: Node | Interface | Link
+        existing_elements: dict = getattr(event.lab, f"_{event.element_type}s")
```
virlos, line 180: Here you're using element_type in a getattr, meaning it must exist. Later, you check for each of the three allowed types, with a warning if it's not one of those. Parhaps a try with AttributeError warning and returning here would be the same future-proofing.

virlos (approved):

tmikuska (approved):

PR #50, opened 2023-08-15
SIMPLE-5729 uploading image files now keeps the extension, removed old deprecated parameter from upload_image_file

daniel-valent:
So I thought there were some inefficiencies in upload_image_file... And when I fixed them, all related test died. So never mind that. Don't touch the Path related things, something about the way they are imported makes them fragile.

_E: Found what the problem was - we are testing Windows paths on Linux, so we have to use PureWindowsPath instead of WindowsPath in tests, which means no interaction with system, which means no Path.exists(). Means that my changes worked fine, but broke tests._


tests/test_image_upload.py
```diff
@@ -74,6 +89,20 @@ def windows_path(path: str):
         yield path
 
 
+# Whenever you modify expected_pass_list, delete files with names from the old list
+# from test_data, set autouse to True and run test_image_upload_file to regenerate files
+@pytest.fixture(autouse=False, scope="module")
```
PatrikMosko, line 94: autouse=False is default btw, no need to explicitly specify it.

tests/test_image_upload.py
```diff
@@ -74,6 +89,20 @@ def windows_path(path: str):
...
+def create_test_files(test_data_path, change_test_dir_module):
```
PatrikMosko, line 95: Is this even used anywhere?

tests/test_image_upload.py
```diff
@@ -61,6 +64,18 @@
 ]
 
 
+@pytest.fixture(scope="module")
```
PatrikMosko, line 67: all of the bellow fixtures should go to conftest.py, so they can be reused by other tests, if there is an assumption that they are too general and might get reused. fixtures switching and creating test_dirs are definitely too. Also, you should go through pytest documentation first, coz pytest already offers everything you are trying to reinvent here, it offers multiple fixtures for creating or switching test_dirs/files.

tests/test_image_upload.py
daniel-valent: Updated the comment above to be clearer that this is not intended to be used normally. 

The fixture was created to initialize the files before every test run and delete them afterwards, but the way parallelization works threw a spanner into that plan. It was easier, in the end, to just create the files once and commit them, and I used this fixture to create them. 

We can also just remove this part, and leave it to those modifying tests to create the files however they want.

tests/test_image_upload.py
daniel-valent: We already have a fixture for this, I just needed it to be module-scoped. Ended up not working the way I wanted, so now I just use said fixture directly.

tests/test_image_upload.py
daniel-valent: See [below.](#discussion_r1298506966)

tests/test_image_upload.py
tmikuska: Why did you need it to be module-scoped?

daniel-valent:
Tests pass, every seems (finally) ready to merge, unless there are further comments.

tests/test_image_upload.py
daniel-valent: Because the fixture using that one was module-scoped, it was module-scoped because function scope was getting applied once each parametrized run, or maybe it was already parallelization problems and I just didn't know that yet and was hoping module scope would solve it? Honestly don't know at this point. Doesn't matter, it had no chance of working in hindsight.

tmikuska: I understand the parallelization pain.  It can be solved by using tmp_path, but it this case it probably makes more sense to pre-create the files as you did.

PR #60, opened 2023-10-11
SIMPLE-5806 PCL support for multiple config files

daniel-valent:


virl2_client/models/node.py
```diff
@@ -404,15 +405,48 @@ def boot_disk_size(self, value: int) -> None:
 
     @property
     def configuration(self) -> str | None:
-        """Return the initial configuration of this node."""
-        self.lab._sync_topology(exclude_configurations=False)
-        return self._configuration
+        """Return the contents of the main configuration file."""
+        self.lab.sync_topology_if_outdated(exclude_configurations=False)
+        return self._configuration[0] if self._configuration else None
```
tmikuska, line 410: doesn't this return a whole config object (not a string)?

virl2_client/models/node.py
```diff
@@ -404,15 +405,48 @@ def boot_disk_size(self, value: int) -> None:
...
+        :param value: The configuration data in one of three formats.
+        """
         self._set_node_property("configuration", value)
-        self._configuration = value
+        self._configuration = (
+            self._configuration if self._configuration is not None else []
+        )
+        if isinstance(value, str):
+            if not self._configuration:
+                self._configuration.append({"name": "Main", "content": value})
```
tmikuska, line 430: shouldn't we get the data from BE instead?  I know that the patch node endpoint returns just node ID, but that can be changed

virl2_client/models/node.py
```diff
@@ -404,15 +405,48 @@ def boot_disk_size(self, value: int) -> None:
...
+                self._configuration[i] = new_configs_by_name[config["name"]]
+
+    @property
+    def configuration_files(self) -> list[dict[str, str]] | None:
+        """
+        Return all configuration files, in a list in the following format:
+        `[{"name": "filename.txt", "content": "<file content>"}]`
+        """
+        self.lab.sync_topology_if_outdated(exclude_configurations=False)
+        return deepcopy(self._configuration)
```
tmikuska, line 449: I'm wondering whether we should have two getters but just one setter?  What if we added a new parameter, e.g., all by default set to False to include all configs in the configuration getter?

virl2_client/models/node.py
daniel-valent: I mean, possible, but then we should probably do that for other properties - not just configs - as to not introduce confusing inconsistency, so I'd rather just force a resync.

virl2_client/models/node.py
daniel-valent: Can't add parameters to a getter; we'd have to change it to a function, which would not be backwards compatible, so for compatibility we'd have to rename the function and deprecate the getter... Yeah, no. Two getters are not really a problem. It's quite clear which returns what, and the setter has pretty clear typing hints to suggest that it accepts multiple types of input, so there should not be any confusion.

virl2_client/models/lab.py
```diff
@@ -245,14 +246,17 @@ def sync_l3_addresses_if_outdated(self) -> None:
 
     @check_stale
     @locked
-    def sync_topology_if_outdated(self) -> None:
+    def sync_topology_if_outdated(self, exclude_configurations=True) -> None:
         """Sync the topology if it is outdated."""
         timestamp = time.time()
-        if (
+        if not exclude_configurations and not self._synced_configs:
```
tmikuska, line 252:
```suggestion
        if not(exclude_configurations or self._synced_configs):
```

virl2_client/models/node.py
```diff
@@ -406,15 +410,48 @@ def boot_disk_size(self, value: int) -> None:
...
+         - or multiple configuration files if passed a list of above dictionaries.
+        Can also use "Main" in place of the filename of the main configuration file.
+
+        :param value: The configuration data in one of three formats.
+        """
         self._set_node_property("configuration", value)
-        self._configuration = value
+        self._configuration = (
+            self._configuration if self._configuration is not None else []
+        )
```
tmikuska, line 432: this can be changed to the following and moved below the `if isinstance(value, str):` block. btw when `self._configuration` can be set to None? for a local node only?
```
if self._configuration is None:
    self._configuration = []
```

virl2_client/models/node.py
```diff
@@ -406,15 +410,48 @@ def boot_disk_size(self, value: int) -> None:
...
+        if isinstance(value, str):
+            if not self._configuration:
+                self._configuration.append({"name": "Main", "content": value})
+            else:
+                self._configuration[0]["content"] = value
```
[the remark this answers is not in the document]
daniel-valent: The main config is *defined* as the first editable config (and the order is retained, all the way through the backend). So it's really the other way around; the first config *is* the main config.

virl2_client/models/node.py
daniel-valent: Oof, brainfart. Thanks.
`Lab.create_node` does:
```python
        for key in ("image_definition", "configuration"):
            if key not in kwargs:
                kwargs[key] = None
```
So when the user does not specify the config when creating the node, and then tries to set config without syncing.

tmikuska (approved):

virl2_client/models/node.py
daniel-valent:
```
         if isinstance(value, str):
             if not self._configuration:
 >               self._configuration.append({"name": "Main", "content": value})
 E               AttributeError: 'NoneType' object has no attribute 'append'
 ```
 ...embarrassing. Will have to reopen this.

tmikuska: ups, sorry about that ... at least I have seen a proof that config can really be None :D

PR #61, opened 2023-10-17
Simple-5953 pcl update for smbios parameters

tulus42:
Updated Node class to accept parameters and added methods for handling node parameters.

### DIFF
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -1326,6 +1326,7 @@ class Lab:
 
         node = self._create_node_local(node_id, **node_data)
         node._state = state
+        node.sync_parameters()
         return node
 
     @locked
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -603,7 +603,7 @@ class Lab:
         if compute_id is not None:
             node._compute_id = compute_id
         self._nodes[node.id] = node
-        node.sync_parameters()
+        self.sync_topology_if_outdated()
         return node
 
     @check_stale
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -24,7 +24,7 @@ import json
 import logging
 import time
 import warnings
-from typing import TYPE_CHECKING, Any, Iterable, Optional
+from typing import TYPE_CHECKING, Any, Iterable
 
 from httpx import HTTPStatusError
 
@@ -36,7 +36,7 @@ from ..exceptions import (
     NodeNotFound,
     VirlException,
 )
-from ..utils import check_stale, locked
+from ..utils import check_stale, get_url_from_template, locked
 from ..utils import property_s as property
 from .cl_pyats import ClPyats
 from .interface import Interface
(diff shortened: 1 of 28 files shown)


virl2_client/models/node.py
```diff
@@ -136,7 +138,7 @@ def __str__(self):
     def __repr__(self):
         return (
             "{}({!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r}, "
-            "{!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r})".format(
+            "{!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r}, {!r})".format(
```
tmikuska, line 141: this needs to be changed, but there's no need to add parameters, please remove them

virl2_client/models/node.py
```diff
@@ -443,6 +446,19 @@ def config(self, value: str) -> None:
         )
         self.configuration = value
 
+    @property
+    def parameters(self) -> dict:
+        self.lab.sync_topology_if_outdated()
+        return self._parameters
+
+    def update_parameters(self, new_params: dict) -> None:
```
tmikuska, line 454: can this be a setter?

virl2_client/models/node.py
```diff
@@ -443,6 +446,19 @@ def config(self, value: str) -> None:
...
+        self._session.patch(self._url_for("node"), json={"parameters": new_params})
+        self.sync_parameters()
+
+    def sync_parameters(self) -> None:
+        node = self._session.get(self._url_for("node")).json()
+        self._parameters = node.get("parameters", {})
```
tmikuska, line 460: I think that this should happen inside `self.lab.sync_topology_if_outdated()`.

virl2_client/models/lab.py
```diff
@@ -597,10 +598,12 @@ def _create_node_local(
             hide_links,
             tags,
             resource_pool,
+            parameters,
         )
         if compute_id is not None:
             node._compute_id = compute_id
         self._nodes[node.id] = node
+        node.sync_parameters()
```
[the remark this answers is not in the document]
tulus42: to retrieve parameters loaded from ND. Otherwise we would have stored locally only custom parameters provided on node creation.

virl2_client/models/node.py
tulus42: I find the setter a little confusing because it's more of a patch than a set. User can update only a few parameters here, not necessarily all of them to rewrite the whole dict.

virlos: I agree that it's not a good fit for a setter.

virl2_client/models/node.py
```diff
@@ -443,6 +446,19 @@ def config(self, value: str) -> None:
...
+    @property
+    def parameters(self) -> dict:
+        self.lab.sync_topology_if_outdated()
+        return self._parameters
+
+    def update_parameters(self, new_params: dict) -> None:
+        self._session.patch(self._url_for("node"), json={"parameters": new_params})
+        self.sync_parameters()
+
+    def sync_parameters(self) -> None:
```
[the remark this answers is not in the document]
virlos: Agreed here.

tmikuska (approved):

PR #64, opened 2023-10-24
SIMPLE-5974 Support for locking computes

daniel-valent:
Added locked_compute_id to node.

### DIFF
--- a/virl2_client/models/node.py
+++ b/virl2_client/models/node.py
@@ -78,6 +78,7 @@ class Node:
         tags: list[str],
         resource_pool: str | None,
         parameters: dict,
+        locked_compute_id: str | None,
     ) -> None:
         """
         A VIRL2 node object representing a virtual machine that serves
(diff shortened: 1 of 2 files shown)


daniel-valent:
Tests still running, will remain in draft until those are done and any potential bugs fixed.

virl2_client/models/node.py
```diff
@@ -495,6 +499,18 @@ def resource_pool(self) -> str:
         self.lab.sync_operational_if_outdated()
         return self._resource_pool
 
+    @property
+    def locked_compute_id(self) -> str | None:
+        """Return the ID of the compute this node is locked to."""
+        self.lab.sync_operational_if_outdated()
```
tmikuska, line 505: I think this should go to `self.node.lab.sync_topology_if_outdated()` as it cannot be changed why a node is running (`compute_id` is the operational property, but that's already covered).

daniel-valent: It can't. It's not *in* the topology. The only sync function that actually retrieves node data directly from the `{lab}/nodes/{id}/` endpoint (the only place which contains `locked_compute_id`) is `Node.sync_operational` (as a side effect of getting operational data) so I just hijacked that.

tmikuska (approved):

PR #68, opened 2023-11-15
updated pyproject.toml and poetry.lock; versions: poetry-1.4, lock for…

PatrikMosko:
time for poetry 1.4+ has come!!!


PatrikMosko:
lock generated with poetry 1.4.0
command: "poetry lock", therefore packages has been upgraded as well.

PR #73, opened 2023-12-05
SIMPLE-6008 PCL backwards compatibility with CML 2.4

tmikuska:


virl2_client/models/users.py
```diff
@@ -36,8 +36,9 @@ class UserManagement:
         "user_id": "users/{username}/id",
     }
 
-    def __init__(self, session: httpx.Client) -> None:
+    def __init__(self, session: httpx.Client, is_rp_supported: bool = True) -> None:
         self._session = session
+        self._is_rp_supported = is_rp_supported
```
virlos, line 41: wouldn't it be swell if the classes shared some context object, where such flags could be stored?

virl2_client/models/users.py
```diff
@@ -106,8 +107,9 @@ def create_user(
             "description": description,
             "admin": admin,
             "groups": groups or [],
-            "resource_pool": resource_pool,
         }
+        if self._is_rp_supported:
+            data["resource_pool"] = resource_pool
```
virlos, line 112: so we assume the value is None then (as no RP could even be created with that server).

virl2_client/models/users.py
```diff
@@ -144,7 +146,7 @@ def update_user(
             data["groups"] = groups
         if password_dict is not None:
             data["password"] = password_dict
-        if resource_pool is not UNCHANGED:
+        if resource_pool is not UNCHANGED and self._is_rp_supported:
```
virlos, line 149: well, let's hope it's None then.

virl2_client/virl2_client.py
```diff
@@ -276,7 +276,9 @@ def __init__(
         self.licensing = Licensing(
             self._session, is_cert_deprecated=controller_version >= Version("2.7.0")
         )
-        self.user_management = UserManagement(self._session)
+        self.user_management = UserManagement(
+            self._session, is_rp_supported=controller_version >= Version("2.5.0")
```
virlos, line 280: RP management itself does not have this flag...

virlos (approved):

tmikuska:
decided not to include this change

PR #74, opened 2023-12-06
SIMPLE-6087 creating a lab no longer raises a warning

daniel-valent:


virl2_client/models/lab.py
```diff
@@ -1188,11 +1188,14 @@ def _import_lab(self, topology: dict) -> None:
         lab_dict = topology.get("lab")
 
         if lab_dict is None:
-            warnings.warn(
-                "Labs created in older CML releases (schema version 0.0.5 or lower) "
-                "are deprecated. Use labs with schema version 0.1.0 or higher.",
-                DeprecationWarning,
-            )
+            # If we just created the node, we skip the warning, since the
```
virlos, line 1191: created the lab. Shouldn't lab create then use the new format maybe?

virlos (approved):

virl2_client/models/lab.py
daniel-valent: Oops, thanks. It's the POST lab endpoint that returns data in this format, and while I could re-package the data, I see no benefit in doing that.

tmikuska: looking at the endpoints it's not just POST lab endpoint that returns a different format than is returned in the export endpoint
while I agree that we should make the responses consistent, I don't see this happening for 2.7

the deprecation warning made sense as long it was used for the import endpoint only
and I personally would add a new attribute `import: bool = True`  and log the warning only when it's True

but most importantly, are you sure that the attribute contains underscore, from what I saw in UI, it's just `created` without underscore

virl2_client/virl2_client.py
```diff
@@ -721,6 +721,7 @@ def create_lab(
             wait_time=self.convergence_wait_time,
             resource_pool_manager=self.resource_pool_management,
         )
+        result["_created"] = True
```
tmikuska, line 724:
```suggestion
        # This is just to skip a deprecation warning in _import_lab
        result["_created"] = True
```

PR #75, opened 2023-12-14
SIMPLE-6105 node.remove no longer always raises NodeNotFound

daniel-valent:
Also omitted the rather useless `owner` argument. Only added it in the first place because `property.__get__` returned it, but it was completely unnecessary in hindsight.


virlos (approved):

PatrikMosko:
For bugs like these, you would ideally also wanted to write a regression test to make sure it wont happen again.

PatrikMosko (approved):

PR #79, opened 2024-01-24
SIMPLE-6204 rename locked compute to pinned

tulus42:
Rename locking compute to pinning compute

### DIFF
--- a/virl2_client/models/node.py
+++ b/virl2_client/models/node.py
@@ -79,7 +79,7 @@ class Node:
         tags: list[str],
         resource_pool: str | None,
         parameters: dict,
-        locked_compute_id: str | None,
+        pinned_compute_id: str | None,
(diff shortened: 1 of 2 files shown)


virl2_client/models/node.py
```diff
@@ -104,7 +104,7 @@ def __init__(
         :param tags: A list of tags associated with the node.
         :param resource_pool: The ID of the resource pool if the node is part
             of a resource pool.
-        :param locked_compute_id: The ID of the compute this node is locked to.
+        :param pinned_compute_id: The ID of the compute this node is locked to.
```
tmikuska, line 107:
```suggestion
        :param pinned_compute_id: The ID of the compute this node is pinned to.
```

tmikuska (changes requested):

PR #85, opened 2024-02-20
SIMPLE-6124 annotations support

PatrikMosko:
Added basic annotations support.
---

**Most notable additions:**
Lab.annotations()
Lab.get_annotation_by_id()
Lab.create_annotation()
Lab.remove_annotations()
Lab.remove_annotation()

**Updated functionalities:**
Lab.update_lab()
Lab.import_lab()

**Other changes:**
 * removed functools.total_ordering from Element objects
 * improved internal documentation (docstrings)

### DIFF
--- a/virl2_client/models/annotation.py
+++ b/virl2_client/models/annotation.py
@@ -123,7 +123,7 @@ ANNOTATION_PROPERTIES_DOCSTRINGS = {
     "y2": "y2 coordinate",
     "z_index": "Z layer (depth) of an annotation",
 }
-ANNOTATION_TYPES = ["text", "line", "ellipse", "rectangle"]
+_ANNOTATION_TYPES = ["text", "line", "ellipse", "rectangle"]
 
 
 class Annotation:
@@ -132,7 +132,12 @@ class Annotation:
         "annotation": "labs/{lab_id}/annotations/{annotation_id}",
     }
 
-    def __init__(self, lab: Lab, annotation_id: str, annotation_type: str) -> None:
+    def __init__(
+        self,
+        lab: Lab,
+        annotation_id: str,
+        annotation_type: AnnotationTypeString,
+    ) -> None:
         """
         A VIRL2 lab annotation.
 
@@ -158,13 +163,14 @@ class Annotation:
             setattr(self, f"_{ppty}", ppty_default)
 
     def __str__(self):
-        return f"Annotation: {self._id}{' (STALE)' if self._stale else ''}"
+        return f"{self.__class__.__name__}: {self._id}{' (STALE)' if self._stale else ''}" 
 
     def __repr__(self):
-        return "{}({!r}, {!r})".format(
+        return "{}({!r}, {!r}, {!r})".format(
             self.__class__.__name__,
             str(self._lab),
             self._id,
+            self._type,
         )
 
     def __eq__(self, other: object):
@@ -220,7 +226,7 @@ class Annotation:
         """
         Checks if given property is recognized by selected annotation type
         """
-        assert annotation_type in ANNOTATION_TYPES
+        assert annotation_type in _ANNOTATION_TYPES
         assert _property in ANNOTATION_PROPERTY_MAP
         annotation_map = {
             "text": 0b1000,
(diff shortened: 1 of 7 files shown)


PatrikMosko:
Lab.update_lab() and Lab.import_lab() is currently the only untested part of the PR.

virl2_client/models/lab.py
```diff
@@ -617,7 +658,7 @@ def remove_node(self, node: Node | str, wait: bool | None = None) -> None:
         """
         Remove a node from the lab.
 
-        If you have a node object, you can also simply do::
```
tmikuska, line 620: I don't think this was a typo.

tmikuska:
is there a simple branch with some annotation tests?

PatrikMosko:
Sadly there isn't, not sure if you mean unit tests or any tests...
We only have API annotation tests, possibly some UI tests.

virl2_client/models/lab.py
PatrikMosko: I will revert it then, after you finish review.

virl2_client/models/lab.py
```diff
@@ -868,6 +950,50 @@ def _create_interface_local(
...
         return iface
 
+    @check_stale
+    @locked
+    def create_annotation(self, annotation_type: str, **kwargs) -> Annotation:
+        """
+        Creates a lab annotation
+
+        :param type: type of annotation (rectangle, ellipse, line or text)
+        :returns: the created annotation
```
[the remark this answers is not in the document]
PatrikMosko: Thanks god we updated that contributing.md so everyone knows.

virl2_client/models/lab.py
```diff
@@ -868,6 +949,59 @@ def _create_interface_local(
...
+        Create a lab annotation
+
+        :param type: type of annotation (rectangle, ellipse, line or text)
+        :returns: the created annotation
+        """
+        url = self._url_for("annotations")
+
+        # create POST json with default annotation property values
+        # override some values by new, expected ones
+        annotation_data = Annotation.get_default_property_values(annotation_type)
```
tmikuska, line 965: Let's server provide the defaults.  If that's not possible, we should really consider changing the API.  It makes no sense to require defaults (and probably not just in this API endpoint).

tmikuska (changes requested):

virl2_client/models/lab.py
PatrikMosko: Every property for each ann. type is required therefore when creating new annotation, you have to provide all of its properties.

I created map of default values, which are the same default values than the one you get when creating annotation through ui.

I made this so you dont have to insert all (required) property values when creating new annotation through PCL, so now you can just do: lab.create_annotation("rectangle")

rschmied: i'd prefer to have a set of functions which defines the properties for annotation, not defined on the backend but locally in the PCL. This is about abstracting complexity and making the use of the API easier by providing "quality of life" functions for the user of the PCL.

1. `Annotation.get_default_properties(annotation_type)` as Patrik proposes which define sane defaults
2. `Annotation.set_default_properties(annotation_type, defaults)` which provides additional values the user wants for this annotation type, they will be merged with the default properties
3. `Annotation.clear_properties(annotation_type)` which will clear the user properties

When the user then creates annotations, the union of the defaults and the user provided properties and the create function are passed to the backend API call to have a complete set of attributes/properties

Example

"defaults for rect": { bg_color="#FFFFFFFF", border_color="#808080FF",  border_style="solid", border_thickness="1", corner_radius=0, layer=0 }

user calls Annotation.create_rect(0,0, 100, 100)  -> properties taken from defaults
user calls Annotation.create_rect(0,0, 100, 100, border_color="#ff000000") -> red rectangle
user calls
Annotations.set_default_properties(Rectangle, border_color="#00ff0000")
Annotations.create_rect(0,0, 100,100)  -> blue rectangle
Annotations.create_rect(100,100, 100,100)  -> another blue rectangle
Annotations.clear_default_properties(Rectangle)
Annotations.create_rect(200,200, 100,100) -> rectangle is black again

This way, the actual creation ALWAYS provides a complete set of defaults and it's easy for the user to control the properties of annotations without having to repeat themselves for every object created.  And the backend does (and should) not have any say in what is a default... the API wants and gets a complete set of properties for every annotation created.

virl2_client/models/annotation.py
```diff
@@ -277,17 +354,20 @@ def _remove_on_server(self) -> None:
     @check_stale
     @locked
     def update(
-        self,
-        annotation_data: dict[str, Any],
-        push_to_server: bool = False,
+        self, annotation_data: dict[str, Any], push_to_server: bool = True
```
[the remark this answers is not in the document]
rschmied: If this is true (and I think it is) then the entire "offline/async" mode should be removed from the PCL as it makes things overly complicated for little to no gain as people are likely not using it in this async mode but rather expect things to be synchronous.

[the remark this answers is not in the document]
tmikuska: raised a new ticket for this

virl2_client/models/lab.py
tmikuska: the reset to defaults is clearly something that we don't do in PCL (we don't even have such options in CML maybe except of sync from a node def)

we do not reset annotations to default in UI either

do we really think that this would be useful?  why don't we just delete and create a new annotation instead?

PatrikMosko: While I like the idea of having something like "default properties overlay", in reality its not really bringing that much of a QoL change.
While now you have to always provide your custom defaults as a second argument to "create" method (Lab.create_annotation("rectangle", user_defined_defaults)), with your proposal in-place, it would be only one (Lab.create_annotation("rectangle")).

I dont think saving one dict var with user-defined defaults and sending one extra parameter in each create call is that much of a pain, sometimes more abstraction hurts.

PatrikMosko: Another issue is we run async, which means there's a miniscule chance that before you send the request with expected defaultsA, other method might update these class-level defaults to defaultsB and they will be used for the first call, expecting defaultsA, too.

Therefore sending the (user-defined subset of) defaults explicitly, while not being overly painful, gives you more control over what is happening.

PR #97, opened 2024-04-29
SIMPLE-6506 removed/deprecated offline mode leftovers

daniel-valent:


tulus42:
I guess when we call it "deprecated" it should still work for a while. Handling removed parameter to tell users it was removed is not a bad idea, but maybe don't call it deprecated.
Or remove it completly and make it clear in the docs/release notes.

tulus42:
I guess when we call it "deprecated" it should still work for a while. Handling removed parameter to tell users it was removed is not a bad idea, but maybe don't call it deprecated.
Or remove it completly and make it clear in the docs/release notes.

PatrikMosko (approved):

PR #106, opened 2024-07-01
SIMPLE-6615 custom mac addresses implementation

daniel-valent:
Still some debugging to do, but opening a draft


virl2_client/models/interface.py
```diff
@@ -385,3 +402,41 @@ def is_connected(self):
...
+        Set a property of the interface.
+
+        :param key: The key of the property to set.
+        :param val: The value to set.
+        """
+        _LOGGER.debug(f"Setting node property {self} {key}: {val}")
+        self._set_interface_properties({key: val})
+
+    @check_stale
+    def _set_interface_properties(self, node_data: dict[str, Any]) -> None:
```
MikuskaPT, line 435: interface_data

virl2_client/models/interface.py
```diff
@@ -144,6 +147,20 @@ def physical(self) -> bool:
...
+    @property
+    def mac_address(self) -> str | None:
+        """Return the MAC address set to the interface.
+        This is the address that will be used when the device is started."""
+        self.node.lab.sync_topology_if_outdated()
+        return self._mac_address
+
+    @mac_address.setter
+    @locked
+    def mac_address(self, value: str) -> None:
```
MikuskaPT, line 159: str | None to allow reset?

MikuskaPT:
we already have mac_address in as_dict [ref](https://github.com/CiscoDevNet/virl2-client/pull/106/files#diff-b0a444300c684343263d9e64e5e80b7dfceeed86a15c654e3ae0478d5378b561R279)
we export operational data for other endpoints, so this could be adjusted in a similar way

PR #131, opened 2025-01-22
Simple 7357 advanced group permissions

marpauli-cisco:


marpauli-cisco:
Updated group model for Advanced group permissions

PR #135, opened 2025-02-21
add params to connection function (fixes #134)

### LINKED ISSUE #134: pyats run_command misbehaving
There are two issues with the current implementation of run_command. Let's take the following command as an example:
```
running_config = pylab.run_command(node_label, "show sdwan run", [], [])
```

We are trying to pass an empty list to init_exec_commands and init_config_commands.

Issue 1: the _prepare_params never returns an empty list as a parameter. This is because if statement is false when list is empty.
```
        params = {}
        if init_exec_commands:
            params["init_exec_commands"] = init_exec_commands
        if init_config_commands:
            params["init_config_commands"] = init_config_commands
        return params
```

I think this should be updated too:
```
params = {}
if init_exec_commands is not None:
    params["init_exec_commands"] = init_exec_commands
if init_config_commands is not None:
    params["init_config_commands"] = init_config_commands
return params
```

Issue 2: the parameters are now given to configure and execute, but not to connect.
```
        if pyats_device not in self._connections or not pyats_device.is_connected():
            if pyats_device in self._connections:
                pyats_device.destroy()

            pyats_device.connect(log_stdout=False, learn_hostname=True)
            self._connections.add(pyats_device)
        params = self._prepare_params(init_exec_commands, init_config_commands)
        if configure_mode:
            return pyats_device.configure(command, log_stdout=False, **params)
        else:
            return pyats_device.execute(command, log_stdout=False, **params)
```

The logic should be adjusted to:
```
        params = self._prepare_params(init_exec_commands, init_config_commands)
        if pyats_device not in self._connections or not pyats_device.is_connected():
            if pyats_device in self._connections:
                pyats_device.destroy()

            pyats_device.connect(log_stdout=False, learn_hostname=True, **params)
            self._connections.add(pyats_device)
        
        if configure_mode:
            return pyats_device.configure(command, log_stdout=False, **params)
        else:
            return pyats_device.execute(command, log_stdout=False, **params)
```

tzarski0:
Before this fix, even the specified custom init_config_commands and init_exec_commands are not executed on connect. After this fix the specified custom init_config_commands and init_exec_commands are executed when connecting to the device.

### DIFF
--- a/virl2_client/models/cl_pyats.py
+++ b/virl2_client/models/cl_pyats.py
@@ -175,15 +175,15 @@ class ClPyats:
         except KeyError:
             raise PyatsDeviceNotFound(node_label)
 
+        params = self._prepare_params(
+            init_exec_commands, init_config_commands, **pyats_params
+        )
         if pyats_device not in self._connections or not pyats_device.is_connected():
             if pyats_device in self._connections:
                 pyats_device.destroy()
 
-            pyats_device.connect(log_stdout=False, learn_hostname=True)
+            pyats_device.connect(log_stdout=False, learn_hostname=True, **params)
             self._connections.add(pyats_device)
-        params = self._prepare_params(
-            init_exec_commands, init_config_commands, **pyats_params
-        )
         if configure_mode:
             return pyats_device.configure(command, log_stdout=False, **params)
         else:


tzarski0:
fixes #134

virlos (approved):

PR #138, opened 2025-02-24
Simple 7407 diagnostics in client

tmikuska:


virl2_client/models/lab.py
```diff
@@ -955,16 +955,17 @@ def _create_link_local(
...
     @check_stale
     @locked
-    def connect_two_nodes(self, node1: Node, node2: Node) -> Link:
+    def connect_two_nodes(self, node1: Node, node2: Node, index: int = 0) -> Link:
         """
         Connect two nodes within a lab.
 
         :param node1: The first node object.
         :param node2: The second node object.
+        :param index: An optional starting interface index (default: 0).
```
virlos: on which node ?

virl2_client/models/lab.py
```diff
@@ -955,16 +955,17 @@ def _create_link_local(
...
         :returns: The created link.
         """
-        iface1 = node1.next_available_interface() or node1.create_interface()
-        iface2 = node2.next_available_interface() or node2.create_interface()
+        iface1 = node1.next_available_interface(index) or node1.create_interface()
+        iface2 = node2.next_available_interface(index) or node2.create_interface()
```
virlos: if the caller requests that interface at least `index` is used, then create_interface should also be passed that index.
Also, if you e.g. want to skip the unused xrv9k interfaces, then the next_available_interface function might want to do it based on the interface label. Then, if you want to connect interface 3 and one of the nodes is an external connector, then your current code will work only if the node does not yet have any interface and create_interface is called.

virl2_client/virl2_client.py
```diff
@@ -805,14 +813,31 @@ def join_existing_lab(self, lab_id: str, sync_lab: bool = True) -> Lab:
...
-        Return the controller diagnostic data as a JSON object.
+        Return selected diagnostic data as a JSON object.
 
+        :param categories: List of diagnostic categories to fetch. If None, fetch all.
         :returns: The diagnostic data.
         """
-        url = self._url_for("diagnostics")
-        return self._session.get(url).json()
+        if categories is None:
+            categories = list(DiagnosticCategory)
```
virlos, line 826: perhaps issue a warning that one should select categories. and the categories can be a varargs parameter.

PR #149, opened 2025-05-06
Simple 7758 added client type identifier

vprysiaz:


virl2_client/models/authentication.py
```diff
@@ -181,7 +181,8 @@ def _request(self, *args, **kwargs):
             raise api_error from None
 
 
-def make_session(base_url: str, ssl_verify: bool | str = True) -> httpx.Client:
+def make_session(base_url: str, ssl_verify: bool | str = True,
+                 client_type_identifier: str = None) -> httpx.Client:
```
virlos, line 185: Can we make the client type a property of the client class, which is settable (and updates the headers accordingly)?
Also, there should be a constant default set at class level, so that subclasses can potentially set it themselves. The header should be initialized in the class' init method.

virlos (approved):

PR #163, opened 2025-08-08
SIMPLE-7778 Updated poetry to version 2

tmikuska:
Updated CML version to 2.10

Update python dependencies

### DIFF
--- a/virl2_client/virl2_client.py
+++ b/virl2_client/virl2_client.py
@@ -181,26 +181,8 @@ class ClientLibrary:
     """Python bindings for the REST API of a CML controller."""
 
     # current client version
-    VERSION = Version("2.9.0")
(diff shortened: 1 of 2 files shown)


virl2_client/virl2_client.py
```diff
@@ -645,7 +645,7 @@ def all_labs(self, show_all: bool = False) -> list[Lab]:
     @locked
     def _remove_stale_labs(self):
         """Remove stale labs from the client library."""
-        for lab in list(self._labs.values()):
+        for lab in self._labs.values():
             if lab._stale:
                 self._remove_lab_local(lab)
```
virlos, line 650: so you can now iterate over a dict and edit it? I don't think so.

PR #178, opened 2025-11-27
CMLDEV-655 refactor opt in

tmikuska:
Added telemetry support

### DIFF
--- a/virl2_client/utils.py
+++ b/virl2_client/utils.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 import warnings
 from collections.abc import Callable
 from contextlib import nullcontext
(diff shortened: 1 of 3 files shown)


virl2_client/utils.py
```diff
@@ -56,6 +57,12 @@ def __repr__(self):
 _CONFIG_MODE = "exclude_configurations=false"
 
 
+class OptInStatus(StrEnum):
```
tmikuska, line 60: did I go to far with the enum?

virlos (approved):

virl2_client/utils.py
virlos: it is fine

PR #179, opened 2025-12-04
CMLDEV-23 pyats logger cleaning

virlos:


virl2_client/models/cl_pyats.py
```diff
@@ -358,3 +361,17 @@ def _destroy_device(self, pyats_device: "Device", raise_exc=True) -> None:
...
+    loggers = logging.root.manager.loggerDict
+    try:
+        names = {con.log.name for con in pyats_device.connectionmgr.connections.values()}
+        names.update(name for name in loggers if name.startswith("unicon.terminal_server."))
+    except (AttributeError, TypeError, KeyError):
+        return
+    for name in names:
+        while name.startswith("unicon."):
+            loggers.pop(name, None)
+            name = name.rsplit(".", 1)[0]
```
tmikuska, line 386: Does the pyATS/unicon team plan to resolve this on their side?

tmikuska (approved):

virl2_client/models/cl_pyats.py
virlos: They plan to at least fix them creating log files in PWD. Once they update socketio, we might ask about that as well - though we'll probably need to tell them how.

PR #182, opened 2025-12-19
CMLDEV-727 JWT authentication

marpauli-cisco:
- added ability to authorize ClientLibrary with JWT
- if credentials are not provided, PCL will prompt to choose auth type (JWT or usr/pwd)
- `authok` url changed to `authentication`
- url for `authentication` changed from `auth_test` to `auth`, because we are using also for getting username when authorized with JWT


virl2_client/models/authentication.py
```diff
@@ -71,16 +72,15 @@ def __init__(self, client_library: ClientLibrary):
...
 
     @property
     def token(self) -> str | None:
         """
         Return the authentication token. If the token has not been set, it is obtained
         from the server.
         """
-        if self._token is not None:
-            return self._token
+        if self.client_library.jwtoken:
```
virlos, line 81: well, if it is none we return none, so we can just return it anyway

virl2_client/models/configuration.py
```diff
@@ -120,26 +125,46 @@ def get_configuration(
...
+
+    if not password:
+        password = _get_prop("VIRL2_PASS") or _get_prop("VIRL_PASSWORD")
+
+    if not jwtoken:
+        jwtoken = _get_prop("VIRL2_JWT")
+
+    if jwtoken or username and password:
+        return host, username, password, jwtoken, ssl_verify
+    else:
```
virlos: since if block ends with return, you don't need to place the rest into an else block

virl2_client/models/configuration.py
```diff
@@ -120,26 +125,46 @@ def get_configuration(
...
+        auth_method = input(
+            "Available authentication methods:\n[1] JW Token\n[2] Username/Password\nChoose method: "
+        )
+        try:
+            auth_method = int(auth_method)
```
virlos: did we have interactive input before? it might be some specific classmethod to setup the client interactively when the user calls for it, but we should not need or fall back to it in the normal setup workflow

virl2_client/virl2_client.py
```diff
@@ -244,16 +247,17 @@ def __init__(
         :raises InitializationError: If no URL is provided, authentication fails or host
             can't be reached.
         """
-        url, username, password, cert = get_configuration(
-            url, username, password, ssl_verify
+        url, username, password, jwtoken, cert = get_configuration(
```
virlos: would it be doable to switch to using a ClientConfig instance for this function and similar?

virl2_client/virl2_client.py
```diff
@@ -379,6 +385,12 @@ def _environ_get(
                 value = default
         return value
 
+    def _get_username(self) -> None:
+        if self.username is None:
+            url = self._url_for("auth")
+            user_info = self._session.get(url).json()
+            self.username = user_info.get("username")
```
virlos, line 450: we might want to remember the userid and other attributes as well, and switch user/pass authentication to also use auth_extended.

virlos: the make test auth call can be merged with this get username function based on whether we have a password or not

virl2_client/models/configuration.py
marpauli-cisco: Before, if username and password were not found in config files or provided via init, then user was prompted to provide them. Here, we just gave the user an option to decide, what to use

virl2_client/models/authentication.py
marpauli-cisco: No, if it's None or "" JWT is obtained from the server

marpauli-cisco:
IMHO there is one downside to using ClientConfig for storing configuration - it is NamedTuple, so once created, we are unable to update it without completely rewriting it. Right now, it is not a problem, because we are not altering in anywhere in code, but we have to have it in mind in the future...
also, there are some (11) failing test, that I've missed, so I'm going to fix them and push, in case someone was itching to approve this :rofl:

virl2_client/models/authentication.py
```diff
@@ -107,7 +107,7 @@ def token(self, value: str | None) -> None:
 
         :param value: The value to set as the authentication token.
         """
-        self._token = value
+        self.client_library.token = value
```
tmikuska: is it token or jwtoken?

virl2_client/models/configuration.py
```diff
@@ -120,26 +144,43 @@ def get_configuration(
...
+    if not password:
+        password = _get_prop("VIRL2_PASS") or _get_prop("VIRL_PASSWORD")
+
+    if not jwtoken:
+        jwtoken = _get_prop("VIRL2_JWT")
+
+    if jwtoken or username and password:
+        return ClientConfig(host, username, password, jwtoken, ssl_verify)
+    auth_method = input(
+        "Available authentication methods:\n[1] JW Token\n[2] Username/Password\nChoose method: "
```
tmikuska:
```suggestion
        "Available authentication methods:\n[1] JWT\n[2] Username/Password\nChoose method: "
```

virl2_client/models/configuration.py
```diff
@@ -120,26 +144,43 @@ def get_configuration(
...
+    try:
+        auth_method = int(auth_method)
+    except ValueError:
+        message = "Invalid authentication method selected."
+        raise InitializationError(message)
+    if auth_method not in [1, 2]:
+        message = "Invalid authentication method selected."
+        raise InitializationError(message)
+    if auth_method == 1:
+        if not (jwtoken := input("Please enter your JW Token: ")):
```
tmikuska:
```suggestion
        if not (jwtoken := input("Please enter your JWT: ")):
```

virl2_client/models/authentication.py
marpauli-cisco: jwtoken

virl2_client/models/authentication.py
virlos: are there legitimate ways in which jwtoken is set to the empty string, and we want to return None instead in this property?

virl2_client/models/configuration.py
virlos: could we instead have user input a value, and if it looks like JWT (after a strip) assume it's that, otherwise assume it is a username and ask for password?

virl2_client/models/configuration.py
```diff
@@ -120,26 +144,43 @@ def get_configuration(
...
+            message = "No JW Token provided."
+            raise InitializationError(message)
+        return ClientConfig(host, username, password, jwtoken, ssl_verify)
+    if not (username or (username := input("Please enter your VIRL username: "))):
```
virlos: drop the VIRL?

virl2_client/virl2_client.py
```diff
@@ -914,6 +882,25 @@ def get_lab_list(self, show_all: bool = False) -> list[str]:
         return self._session.get(**url).json()
 
 
+def make_client(client_config: ClientConfig) -> ClientLibrary:
```
virlos: why can't it be a method of ClientConfig? We've been using it that way in testing as well. If you want to move the class by all means (but import it into this module), and keep the API compatible. If there is a circular import issue (how) then that can be surely solved while keeping the public API in place.

virlos (approved):

virl2_client/models/configuration.py
tmikuska: let's do that in CMLDEV-777

virl2_client/models/authentication.py
tmikuska: I don't think that jwtoken can be an empty string unless set by a user manually, and even then it wouldn't work

marpauli-cisco: well, on ClientLibrary init you can specify username, password and JWT as an empty string as well...

virl2_client/virl2_client.py
```diff
@@ -356,6 +358,141 @@ def _make_test_auth_call(self) -> None:
             raise
         except httpx.HTTPError as exc:
             raise InitializationError(exc)
+        user_info = response.json()
+        self.user_id = user_info.get("id")
+        self.admin = user_info.get("admin", False)
```
tmikuska, line 448: do we need user_id and admin?

virl2_client/virl2_client.py
```diff
@@ -356,6 +358,141 @@ def _make_test_auth_call(self) -> None:
...
+        if self.username is None:
+            self.username = user_info.get("username")
```
tmikuska, line 450: I'd overwrite username all the time

virl2_client/virl2_client.py
virlos: I would fetch user_id at least for all auth methods (i.e. when doing user/pass); I recall some model actually wanted to know it at some point, and maybe fetches it on the fly. Knowing whether we are an admin can become useful too.

virlos (approved):

PR #184, opened 2026-01-07
CMLDEV-702 fix pyats console switch

virlos:


virl2_client/models/cl_pyats.py
```diff
@@ -128,7 +135,7 @@ def sync_testbed(self, username: str, password: str) -> None:
         self._testbed = self._load_pyats_testbed(testbed_yaml)
         self.set_termserv_credentials(username, password)
 
-    def switch_pyats_serial_console(self, node_label: str, console_number: int) -> None:
+    def switch_serial_console(self, node_label: str, console_number: int | str) -> None:
```
tmikuska, line 138: I'd allow either int (preferred) or str but not both, but I don't care that much.

virl2_client/models/cl_pyats.py
```diff
@@ -142,23 +149,37 @@ def switch_pyats_serial_console(self, node_label: str, console_number: int) -> N
...
+        any connections. Then either set the password, or path to an identity
+        file if SSH authentication with public keys is set up on the server.
+        By default, this function disables authentication agents and identity
+        files that would be loaded from the environment and running user ssh
+        configuration, so that the passed password or key is attempted first.
+        Pass empty string or custom SSH options to override this behavior.
+        """
         terminal = self._testbed.devices.terminal_server
+        if ssh_options is None:
+            ssh_options = DEFAULT_SSH_OPTIONS
```
tmikuska: I'd move this below (closer to other ssh_options related operations).

virl2_client/models/cl_pyats.py
```diff
@@ -192,6 +223,16 @@ def _reconnect(self, pyats_device: "Device", params: dict) -> None:
             _remove_unicon_loggers(pyats_device)
         self._connections.add(pyats_device)
 
+    def _maybe_reconnect(self, pyats_device: "Device", params: dict):
+        """Reconnect the device and/or its proxy as needed"""
+        # check terminal server first, it may be missing, None or list of connections
+        #if proxies := getattr(pyats_device, "proxy_connections", None):
+        #    for proxy in proxies:
+        #        if not proxy.is_connected or not proxy.spawn.fd:
+        #            proxy.connect()
```
tmikuska: what about this?

virl2_client/models/cl_pyats.py
```diff
@@ -192,6 +223,16 @@ def _reconnect(self, pyats_device: "Device", params: dict) -> None:
             _remove_unicon_loggers(pyats_device)
         self._connections.add(pyats_device)
 
+    def _maybe_reconnect(self, pyats_device: "Device", params: dict):
```
tmikuska: _connect_if_disconnected?  or could this be collapsed into _reconnect?

virl2_client/models/cl_pyats.py
virlos: works the same, int is there because it's logical, but it's stringized anyway. Also we are not even checking if multiple chars are passed which would break the switching logic, but that should be easily avoidable by end-users.

virl2_client/models/cl_pyats.py
virlos: well we don't need to use None here at all in the end version

virl2_client/models/cl_pyats.py
virlos: apparently it works without this check after all, and Device.is_connected checks the proxy state as well.

tmikuska (approved):

PR #188, opened 2026-01-17
CMLDEV-786 - switch to using link ID as capture key

virlos:


virl2_client/models/link.py
```diff
@@ -468,54 +475,41 @@ def capture_status(self) -> dict:
     def capture_key(self) -> str:
         """
         Get the capture key (UUID) for the packet capture on this link.
+        The capture key is deprecated and since 2.10 matches the link uuid.
 
         :returns: The capture key as a string.
-        :raises: HTTP exception if no capture is running on this link.
         """
+        warnings.warn("Use of capture keys is deprecated. Use the link ID instead.")
```
tmikuska: This was implemented just recently, can't we remove it directly instead?

xorrkaz: As someone now using the PCAP key (directly from the API), I hope even if it is removed from the PCL, the endpoint still works.  The dance to maintain backward compatibility isn't something I look forward to taking on.

tmikuska: the endpoint is now deprecated in CML, but it was not implemented in PCL until now and my question is whether it makes sense to introduce a deprecated endpoint

virlos: I guess we can remove it before ever releasing it.

tmikuska (approved):

PR #189, opened 2026-01-18
CMLDEV-727 Revert the old authentication

tmikuska:
… for previous but still supported CML releases

### DIFF
--- a/virl2_client/virl2_client.py
+++ b/virl2_client/virl2_client.py
@@ -304,6 +304,7 @@ class ClientLibrary:
 
     _URL_TEMPLATES = {
         "auth": "authentication",
+        "old_auth": "authok",
         "system_info": "system_information",
         "import": "import",
         "import_1x": "import/virl-1x",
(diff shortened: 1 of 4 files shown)


virl2_client/virl2_client.py
```diff
@@ -443,11 +447,11 @@ def _make_test_auth_call(self) -> None:
...
-        user_info = response.json()
-        self.user_id = user_info.get("id")
-        self.admin = user_info.get("admin", False)
-        if self.username is None:
-            self.username = user_info.get("username")
+        if new_auth:
+            user_info: dict = response.json()
+            self.user_id = user_info["id"]
+            self.admin = user_info.get("admin", False)
+            self.username = user_info["username"]
```
tmikuska, line 454: do we want to throw a deprecation warning for the old url, e.g. when using PCL 2.10 with CML 2.9 or older?  or is it sufficient to throw just the other one (Please ensure the client version is compatible with the controller ...)?

virlos (approved):

PR #190, opened 2026-01-18
CMLDEV-428 make pyats configurable per node image definition

tmikuska:


virl2_client/models/node.py
```diff
@@ -470,6 +480,44 @@ def parameters(self) -> dict:
...
+        This updates the node on the controller with a ``pyats`` field whose
+        structure matches the backend expectation, typically::
+
+            {
+                "username": "<user>",
+                "password": "<pass>",
+                "enable_password": "<enable_password_pass>"
+            }
+        """
+        pyats = self._pyats.copy()
```
virlos, line 516: why can't you edit the single instance but make a copy? What if a script holds onto the old instance?

tmikuska: the instance should be updated locally only after it's updated on the server

virl2_client/models/node.py
```diff
@@ -470,6 +480,44 @@ def parameters(self) -> dict:
...
+        if username is not UNCHANGED:
+            pyats["username"] = username
+        if password is not UNCHANGED:
+            pyats["password"] = password
+        if enable_password is not UNCHANGED:
+            pyats["enable_password"] = enable_password
+        self._set_node_property("pyats", pyats)
```
tmikuska, line 523: this can be no-op, do we want not to send a request in that case?

virlos: send it anyway just to make sure.

virl2_client/models/node.py
```diff
@@ -114,6 +121,9 @@ def __init__(
         self._hide_links: bool = kwargs.get("hide_links", False)
         self._tags: list[str] = kwargs.get("tags", [])
         self._parameters: dict = kwargs.get("parameters", {})
+        self._pyats: dict[str, str | None] = kwargs.get(
+            "pyats", {"username": None, "password": None, "enable_password": None}
```
xorrkaz, line 126: If I read this, the default is None in the constructor, but "UNCHANGED" in the setter.  Shouldn't they agree?

xorrkaz (commented):
Thanks for doing this!  I think the usage matches the use case I have in mind.

virl2_client/models/node.py
tmikuska: you're correct, the type annotations need to be fixed

FYI, we use UNCHANGED when we cannot set None as the default value mostly when None is used to clear the current value

virl2_client/models/node.py
```diff
@@ -114,6 +122,9 @@ def __init__(
         self._hide_links: bool = kwargs.get("hide_links", False)
         self._tags: list[str] = kwargs.get("tags", [])
         self._parameters: dict = kwargs.get("parameters", {})
+        self._pyats: dict[str, str | None] = kwargs.get(
```
xorrkaz, line 125: Hmmm, given that the _set_ values of the credentials are retrievable, why not initialize these to the default values?

virlos: the Node table only holds values that are explicit overrides made by the user, and the node API responses only set the values that have been altered. Otherwise you would not be able to distinguish whether the value is intentionally set by the user or came from the definitions. 

If you changed the node def, the change would not be reflected on the nodes themselves, or it would also change nodes where the previous value was meant to be used by the user. Moreover in case of pyats, we have these also set on image definitions because of the ASAv change a while back. So changing the image def would also need to be reflected in the APIs.

We do create node config entries for each configuration file. There is a lot of (often weird) syncing logic involved with keeping the best version of each config present on the node, and I'd like to avoid doing it for the other attributes as well.

You do get the final values in the pyats testbed response.

xorrkaz: Okay.  There is at least a way to fetch them, I guess at the node def level.  Might be nice, in the future, to have a node detail overlay in the API response that enriches the dynamic node data with node def-level data, too.

tmikuska: That can be easily done, the question is where to put such information.  We have operational data (used data when a node is defined), but it doesn't feel right to put it there as pyats credentials can change after a node is defined.

Currently, UI computes this on it's own.  PCL could do the same, but if we added a new structure to the node response, e.g. inherited which could contain all data from node and image definitions (currently cpus, cpu_limit, data_volume, boot_disk_image, ram, pyats_username, pyats_password, pyats_enable_password), we could save some computation time in both UI and PCL.

(Anyway, I'm going to merge this PR and any other changes would be done in a new PR).

virlos: well, some "resolved" mode flag could compute the actual image and all the information from NodeDefineParameters. It is definitely a 2.11 thing at this point

PR #191, opened 2026-01-19
Moved the deprecation warning for obtaining configuration via input 

tmikuska:
... to correct place

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -20,6 +20,9 @@ _TEST_ENV = {
     "VIRL_USERNAME": "test_admin",
     "VIRL2_PASS": "test_test123",
     "VIRL_PASSWORD": "test_test123",
+    "VIRL2_JWT": "test_jwt_token",
+    "CA_BUNDLE": "/path/to/ca-bundle.pem",
+    "CML_VERIFY_CERT": "1",
 }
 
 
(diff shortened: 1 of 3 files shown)


tests/test_configuration.py
```diff
@@ -79,10 +88,65 @@ def test_home_directory_virlrc(
...
+    monkeypatch.setenv("VIRL2_JWT", _TEST_ENV["VIRL2_JWT"])
+
+    config = ClientConfig.get_configuration(
+        url=None, username=None, password=None, jwtoken=None, ssl_verify=None
+    )
+
+    assert config.url == _TEST_ENV["VIRL2_URL"]
+    assert config.jwtoken == _TEST_ENV["VIRL2_JWT"]
+    assert config.username is None
+    assert config.password is None
```
virlos, line 116: are there any tests that say that e.g. username gets set when the client is made from this config? (somebody could add such tests)

virl2_client/virl2_client.py
```diff
@@ -245,18 +261,8 @@ def get_configuration(
         ssl_verify: bool | str | None,
         allow_inputs: bool | None = None,
     ) -> "ClientConfig":
-        populate_functions = [cls._populate_from_env, cls._populate_from_rc_files]
-        if allow_inputs is None:
+        if emit_warning := allow_inputs is None:
             allow_inputs = sys.stdin.isatty()
```
virlos: currently for compatibility, allow_inputs should default to True so that scripts do get the old behavior as a fallback. this is the new/future behavior IIRC? I may be misremembering

virlos (approved):

tests/test_configuration.py
tmikuska: there's one in another test module, I'm going to move it here and extend it

PR #192, opened 2026-02-07
CMLDEV-689 Implemented RADIUS auth

tmikuska:
CMLDEV-489 Added timeout property to LDAP auth and fixed password setter


tmikuska:
merged since I need this for RC but feel free to comment and I will incorporate all feedback into the next RC

PR #194, opened 2026-02-27
CMLDEV-910 added raise for InitializationError when jwt is incorrect and usernam…

vprysiaz:
…e/password not provided

### DIFF
--- a/virl2_client/models/authentication.py
+++ b/virl2_client/models/authentication.py
@@ -28,7 +28,7 @@ from uuid import uuid4
 
 import httpx
 
-from ..exceptions import APIError, InitializationError
+from ..exceptions import APIError
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


PatrikMosko (approved):

virl2_client/models/authentication.py
```diff
@@ -124,6 +124,12 @@ def auth_flow(
         if response.status_code == 401:
             _LOGGER.warning("re-auth called on 401 unauthorized")
             self.token = None
+            if not (self.client_library.username and self.client_library.password):
+                raise InitializationError(
```
tmikuska: I guess that APIError is not ideal name either, but this is mostly problem on re-auth not on initialization

do we know why and if we need to use APIError when a request fails?

vprysiaz: I choose InitializationError as when it is raised we have good error handling flow. and flag that is responsible for "raise on auth" or something like that. 
for me Init errors looks good. because providing token only without password and calling re-auth is a problem in how pcl was initialized. but feel free to suggest something else

tmikuska:
next time be careful to create a branch from dev

virl2_client/models/authentication.py
tmikuska: I went with APIError as that's what we use in integration tests

PR #195, opened 2026-02-27
CMLDEV-934 Drop deprecated endpoints and fields in CML 2.11 API

tmikuska:


tests/test_client_library.py
```diff
@@ -820,7 +820,7 @@ def test_import_lab_offline_deprecated(
     topology_file_path = test_data_dir / "sample_topology.json"
     with open(topology_file_path) as fh:
         topology_file = fh.read()
-        with pytest.deprecated_call():
+        with pytest.raises(TypeError):
```
tmikuska, line 915: we usually do not test remove code, so this can be removed, but doesn't hurt

virl2_client/virl2_client.py
```diff
@@ -904,19 +901,14 @@ def get_diagnostics(self, *categories: DiagnosticsCategory) -> dict:
...
-                "Use 'ClientLibrary.get_diagnostics(DiagnosticsCategory.ALL)' or "
-                "'ClientLibrary.get_diagnostics(DiagnosticsCategory.COMPUTES, "
-                "DiagnosticsCategory.LABS, ...)' with specific categories instead.",
-                DeprecationWarning,
+            raise ValueError(
+                "No diagnostics category provided. Use "
+                "ClientLibrary.get_diagnostics(DiagnosticsCategory.ALL) "
+                "or provide one or more explicit categories."
             )
-            categories = [DiagnosticsCategory.ALL]
```
tmikuska, line 919: this seems a little harsh, but displaying what we usually do

I'd prefer to make ALL the default option instead even though it doesn't match API behavior

virlos: no we wanted to avoid people being lazy and asking for everything every time. they need to think which values they actually need

virlos (approved):

PR #197, opened 2026-03-07
Fix bugs and apply improvements across client library

tmikuska:
- Fix PyatsException to inherit from VirlException instead of Exception
- Fix ControllerNotFound to use __init__ instead of class attribute for message
- Fix ControllerNotFound instantiation (raise ControllerNotFound())
- Fix sync_states staleness handling for nodes, interfaces, and links
- Fix _import_link parameter order (iface_a_id before iface_b_id)
- Fix incorrect log message in Interface._set_interface_property
- Refactor Annotation validation using _VALID_KEYS frozensets
- Refactor Version comparisons using _as_tuple helper
- Refactor get_lab_list to pass params directly to session.get
- Optimize Node.next_available_interface with direct slicing

### DIFF
--- a/virl2_client/models/annotation.py
+++ b/virl2_client/models/annotation.py
@@ -21,7 +21,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, Union
 
 from ..exceptions import InvalidProperty
 from ..utils import check_stale, get_url_from_template, locked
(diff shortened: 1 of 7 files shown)


virl2_client/models/annotation.py
```diff
@@ -449,9 +465,8 @@ def _update(self, annotation_data: dict[str, Any], push_to_server: bool) -> None
             raise ValueError("Can't change annotation type.")
 
         # make sure all properties we want to update are valid
-        existing_keys = dir(self)
         for key in annotation_data:
-            if key not in existing_keys:
+            if key not in self._VALID_KEYS:
```
tmikuska, line 469: this was super implicit/magical, but I can bring it back cause new changes are extensive

virl2_client/models/lab.py
```diff
@@ -1749,16 +1768,16 @@ def _handle_import_annotations(self, topology: dict) -> None:
     def _import_link(
         self,
         link_id: str,
-        iface_b_id: str,
         iface_a_id: str,
+        iface_b_id: str,
```
tmikuska, line 1772: not sure why this was done this way, but this shouldn't need any deprecation

virl2_client/models/annotation.py
virlos: it's fine

virlos (approved):

PR #198, opened 2026-03-07
Convert relative imports to absolute imports

tmikuska:
Replace all relative imports (from . / from ..) with absolute imports (from virl2_client.) across the non-test codebase for consistency.

### DIFF
--- a/virl2_client/__init__.py
+++ b/virl2_client/__init__.py
@@ -19,8 +19,13 @@
 #
 """Python client library for the Cisco Modeling Labs (CML) REST API."""
 
-from .exceptions import InterfaceNotFound, LabNotFound, LinkNotFound, NodeNotFound
-from .virl2_client import ClientConfig, ClientLibrary, InitializationError
+from virl2_client.exceptions import (
+    InterfaceNotFound,
+    LabNotFound,
+    LinkNotFound,
+    NodeNotFound,
+)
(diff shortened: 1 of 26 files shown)


tmikuska:
let me know what you think

virlos:
not sure what tools suggest this over relative imports. relative imports are a feature, i.e. not having to use the project name everywhere, and make moving packages easier, while also showing that the import is local. so, I find this unnecessary, though I also do not object against changing it.

tmikuska:
> not sure what tools suggest this over relative imports. relative imports are a feature, i.e. not having to use the project name everywhere, and make moving packages easier, while also showing that the import is local. so, I find this unnecessary, though I also do not object against changing it.

In the main code base, this is more of an exception in pkgtool and tests, but here relative imports represent 1/3 of imports.  I just wanted to hear everyone's opinion before making any change.  I prefer absolute imports over relative ones, but I have no strong feelings and the main reason why I raised this this way is the majority of imports are still absolute.

PatrikMosko:
I agree with @virlos , relative imports have their place and there's no reason to enforce consistency on this front.

It might make sense to remove certain relative imports especially the ".."-ones, but not all... also, there's a good chance the "virl2_client" will become "cml2_client" or something, you know where I am going with this

Issue #153, opened 2025-06-13
Performance optimization when querying many labs/nodes


sgherdao:
Hi,

I’ve noticed some performance issues when querying a controller with a large number of labs and nodes. For example, running the cmlutils `cml ls` command on a controller with ~40 labs takes approximately 26 seconds:

```
cml ls  1.18s user 0.23s system 5% cpu 26.219 total
```

When enabling logging, I can see redundant calls to the users, and topology endpoints.

```
2025-06-13 19:53:56,501 - httpx - INFO - HTTP Request: GET https://cml/api/v0/system_information "HTTP/1.1 200 OK"
2025-06-13 19:53:57,042 - httpx - INFO - HTTP Request: POST https://cml/api/v0/authenticate "HTTP/1.1 200 OK"
2025-06-13 19:53:57,120 - httpx - INFO - HTTP Request: GET https://cml/api/v0/authok "HTTP/1.1 200 OK"
2025-06-13 19:53:57,184 - httpx - INFO - HTTP Request: GET https://cml/api/v0/system_information "HTTP/1.1 200 OK"
2025-06-13 19:53:57,400 - httpx - INFO - HTTP Request: GET https://cml/api/v0/system_information "HTTP/1.1 200 OK"
2025-06-13 19:53:57,869 - httpx - INFO - HTTP Request: POST https://cml/api/v0/authenticate "HTTP/1.1 200 OK"
2025-06-13 19:53:57,945 - httpx - INFO - HTTP Request: GET https://cml/api/v0/authok "HTTP/1.1 200 OK"
2025-06-13 19:53:58,027 - httpx - INFO - HTTP Request: GET https://cml/api/v0/users "HTTP/1.1 200 OK"

# Querying labs' info

2025-06-13 19:53:58,109 - httpx - INFO - HTTP Request: GET https://cml/api/v0/labs "HTTP/1.1 200 OK"

2025-06-13 19:53:58,228 - httpx - INFO - HTTP Request: GET https://cml/api/v0/labs/73060ff2-a0c4-4475-9f55-e62401e2e5e2/topology "HTTP/1.1 200 OK"
2025-06-13 19:53:58,435 - httpx - INFO - HTTP Request: GET https://cml/api/v0/users "HTTP/1.1 200 OK"
[...]

PR #199, opened 2026-03-07
Store users locally to improve performance when querying labs/nodes

tmikuska:
Fixes https://github.com/CiscoDevNet/virl2-client/issues/153

### DIFF
--- a/virl2_client/models/user.py
+++ b/virl2_client/models/user.py
@@ -20,6 +20,7 @@
 
 from __future__ import annotations
 
+import time
 import warnings
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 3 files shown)


virl2_client/models/user.py
```diff
@@ -58,6 +68,36 @@ def users(self) -> list[dict]:
         url = self._url_for("users")
         return self._session.get(url).json()
 
+    def sync_users(self) -> None:
+        """Fetch all users from the server and store them locally."""
+        user_list = self.users()
+        self._users_by_id = {user["id"]: user for user in user_list}
+        self._last_sync_users_time = time.time()
+
+    def sync_users_if_outdated(self) -> None:
```
PatrikMosko, line 77: ok. I disliked this design before, and I dislike it now as well. I feel like this is only warranted in cases where sync method requires some additional stuff to do, if its a simple TTL mechanism, I feel we would be better off with memoization/cache.

PatrikMosko: I think using TTLCache (from cachetools) or similar memoization object should be sufficient... we can create wrappers and separate TTLCache for each feature that would require its own cache and set TTL to 1+ seconds

```
user_cache = cachetools.TTLCache(maxsize=1, ttl=1) 
...
class user_management:

    @cached(cache=user_cache)
    def users(self):
        url = self._url_for("users")
        return self._session.get(url).json()

    def sync_users(self):
        user_cache.clear()
```

its just an idea... implementing same method for dozen of managers/features seems very redundant

tmikuska: I don't like it either but it works and I just went with what is already proven to work.  This could be refactored but it should be done across the code and imo it's not a priority.

PR #209, opened 2026-04-07
CMLDEV-923 ruff lint fixes

tmikuska:
An initial batch of ruff fixes.  I can split commits in individual PRs if needed.

### DIFF
--- a/virl2_client/models/annotation.py
+++ b/virl2_client/models/annotation.py
@@ -507,7 +507,7 @@ class Annotation:
     @check_stale
     def _remove_on_server(self) -> None:
         """Remove annotation on the server side."""
-        _LOGGER.info(f"Removing annotation {self}")
+        _LOGGER.info("Removing annotation %s", self)
         url = self._url_for("annotation")
         self._session.delete(url)
 
@@ -555,7 +555,7 @@ class Annotation:
         :param key: The name of the property to set.
         :param val: The value to set.
         """
-        _LOGGER.debug(f"Setting annotation property {self} {key}: {val}")
+        _LOGGER.debug("Setting annotation property %s %s: %s", self, key, val)
         self._set_annotation_properties({key: val})
 
     @check_stale
--- a/virl2_client/models/annotation.py
+++ b/virl2_client/models/annotation.py
@@ -21,7 +21,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Any, Literal, Union
+from typing import TYPE_CHECKING, Any, Literal
 
 from ..exceptions import InvalidProperty
 from ..utils import check_stale, get_url_from_template, locked
@@ -33,13 +33,12 @@ if TYPE_CHECKING:
     from .lab import Lab
 
     AnnotationTypeString = Literal["text", "line", "ellipse", "rectangle"]
-    AnnotationType = Union[
-        "Annotation",
-        "AnnotationRectangle",
-        "AnnotationEllipse",
-        "AnnotationLine",
-        "AnnotationText",
-    ]
+    AnnotationType = (
+        "AnnotationRectangle"
+        | "AnnotationEllipse"
+        | "AnnotationLine"
+        | "AnnotationText"
+    )
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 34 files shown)


tests/test_system_lab_repositories.py
```diff
@@ -81,7 +81,12 @@ def system_with_repos(
     """
     lab_repo_mgmt = mock_lab_repository_management
     for repo_data in MOCK_LAB_REPOSITORIES_LIST:
-        lab_repo_mgmt.add_lab_repository_local(**repo_data)
+        lab_repo_mgmt.add_lab_repository_local(
```
virlos: so if something gets added to repo_data, this will have to be also updated. Sometimes I don't know about those rules people.

virl2_client/models/annotation.py
```diff
@@ -33,13 +33,13 @@
...
     AnnotationTypeString = Literal["text", "line", "ellipse", "rectangle"]
-    AnnotationType = Union[
-        "Annotation",
-        "AnnotationRectangle",
-        "AnnotationEllipse",
-        "AnnotationLine",
-        "AnnotationText",
-    ]
+    AnnotationType = (
+        "Annotation"
```
virlos: should the base annotation be included in such a union?

virl2_client/models/lab_repository.py
```diff
@@ -81,7 +81,7 @@ def _url_for(self, endpoint: str, **kwargs: str) -> str:
         return get_url_from_template(endpoint, self._URL_TEMPLATES, kwargs)
 
     @property
-    def id(self) -> str:  # noqa: A003
+    def id(self) -> str:
```
virlos, line 84: you may as well rename this and then have id = repo_id.

virl2_client/models/lab_repository.py
```diff
@@ -258,22 +271,22 @@ def refresh_lab_repositories(self) -> dict[str, dict[str, Any]]:
 
     def add_lab_repository_local(
         self,
-        id: str,
+        repo_id: str,
```
virlos: (what if someone already used explicit arg passing in their script)

virl2_client/models/link.py
```diff
@@ -319,20 +319,22 @@ def wait_until_converged(
         :raises RuntimeError: If the link does not converge within the specified number
             of iterations.
         """
-        _LOGGER.info(f"Waiting for link {self.id} to converge")
+        _LOGGER.info("Waiting for link %s to converge", self.id)
```
virlos: maybe using _id would help

virl2_client/models/node_image_definition.py
```diff
@@ -329,17 +327,15 @@ def print_progress_bar(
     :param length: The length of the progress bar.
     """
     percent = f"{100 * (cur / float(total)):.1f}"
-    filled_len = int(round(length * cur / float(total)))
+    filled_len = round(length * cur / float(total))
     bar = "#" * filled_len + "-" * (length - filled_len)
     raw_elapsed = time.time() - start_time
     elapsed = time.strftime("[%H:%M:%S]", time.gmtime(raw_elapsed))
-    print(
```
virlos, line 336: (some prints are actually fine no? Do rules actually say using sys.stdout is the right thing to do?

virlos (approved):

tests/test_system_lab_repositories.py
tmikuska: I dropped this for now

virl2_client/models/lab_repository.py
tmikuska: this is an addition in 2.10 but I need to come up with something for notices

virl2_client/models/node_image_definition.py
tmikuska: reverted

virl2_client/models/annotation.py
```diff
@@ -33,13 +33,12 @@
...
-        "AnnotationEllipse",
-        "AnnotationLine",
-        "AnnotationText",
-    ]
+    AnnotationType = (
+        "AnnotationRectangle"
+        | "AnnotationEllipse"
+        | "AnnotationLine"
+        | "AnnotationText"
+    )
```
tmikuska, line 41: this is a battle between SQ and ruff ... which one do we prefer?

virlos: I prefer the new style

virlos (approved):

PR #210, opened 2026-04-08
Rename `id` parameters, fix LabRepository ownership, and protect object IDs

tmikuska:
- Rename `id` to `notice_id` in SystemNotice and add_system_notice_local (A002)
- Rename `id` to `repo_id` in LabRepository and add_lab_repository_local (A002)
- Change LabRepository to reference its owning LabRepositoryManagement instead of SystemManagement, removing broken hasattr fallback branches
- Prevent _update from overwriting object IDs in Node, Interface, ResourcePool, ComputeHost, and SystemNotice
- Update folder type to str | None in LabRepository

### DIFF
--- a/virl2_client/models/system.py
+++ b/virl2_client/models/system.py
@@ -219,14 +219,13 @@ class SystemManagement:
         for compute_host in compute_hosts:
             compute_host.pop("nodes", None)  # removed in 2.10
             compute_id = compute_host.pop("id")
-            compute_host["compute_id"] = compute_id
             if compute_id in self._compute_hosts:
                 self._compute_hosts[compute_id]._update(
                     compute_host, push_to_server=False
                 )
             else:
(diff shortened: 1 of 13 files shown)


virl2_client/models/system.py
```diff
@@ -387,18 +386,18 @@ def add_compute_host_local(
 
     def add_system_notice_local(
         self,
-        id: str,
+        notice_id: str,
```
tmikuska, line 389: I found deprecation to be unnecessary annoying and I'd rather take the bullet here (lab_repo is a new feature in 2.10)

virlos: fine. you can leave an id = notice_id after the property is created, and usage of named attributes by customer scripts is not that likely.

virlos (approved):

PR #213, opened 2026-04-09
CMLDEV-779: Updated _import_lab and _set_owner; Updated failing tests;

BregaladTaran:
Resolves [CMLDEV-779](https://cisco-learning.atlassian.net/browse/CMLDEV-779).

- Adds `Lab.owner_id` property and persists `_owner_id` from `_import_lab`,
  `update_lab_properties`, and `_set_owner`.
- `_set_owner` now records both the resolved username and the user id,
  drops the broken `try/except`, and no longer wipes `_owner_id` after a
  successful resolution.
- Tests updated to assert the new `owner_id` value alongside `owner`.

Follow-up (not in this PR):
- Replace `_set_owner` with an `owner` property setter (per @PatrikMosko's
  review).
- Unify the two `_import_lab` data paths once the backend confirms a
  single response shape.

### DIFF
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -1679,8 +1679,12 @@ class Lab:
 
             # API returns both "owner" (ID) and "owner_username"
             owner_id = topology.get("owner") or topology.get("lab_owner")
-            owner_username = topology.get("owner_username", default_owner)
-            self._set_owner(owner_id, owner_username)
+            owner_username = topology.get("owner_username")
+            if owner_username:
+                self._owner = owner_username
+                self._owner_id = owner_id
+            else:
+                self._set_owner(owner_id, default_owner)
 
             if autostart := topology.get("autostart"):
                 self._autostart = autostart
@@ -2211,7 +2215,11 @@ class Lab:
         if "owner" in properties:
             owner_id = properties.get("owner")
             owner_username = properties.get("owner_username")
-            self._set_owner(owner_id, owner_username)
+            if owner_username:
+                self._owner = owner_username
+                self._owner_id = owner_id
+            else:
+                self._set_owner(owner_id)
 
         self._autostart.update(properties.get("autostart", {}))
 
@@ -2481,20 +2489,18 @@ class Lab:
         self, user_id: str | None = None, user_name: str | None = None
     ) -> None:
         """
-        Sets the owner ID and username. If user_id is provided, fetches the user
-        information directly from the API to get the username. Falls back to the
-        provided user_name if the ID is None or the user doesn't exist.
+        Sets the owner ID and username. If user_id is provided, uses the cached
+        user list to resolve the username efficiently. Falls back to the
+        provided user_name if the ID is None or the user doesn't exist in cache.
 
         :param user_id: User unique identifier.
-        :param user_name: Username.
+        :param user_name: Username to use as fallback.
         """
         if user_id:
-            try:
-                user_info = self._user_management.get_user(user_id)
-                user_name = user_info.get("username")
-                self._owner_id = user_id
-            except Exception:
-                self._owner_id = user_id if user_name else None
+            resolved = self._user_management.get_username(user_id)
+            if resolved is not None:
+                user_name = resolved
+            self._owner_id = user_id
         else:
             self._owner_id = None
         self._owner = user_name
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -1659,14 +1659,13 @@ class Lab:
         Replace lab properties with the given topology.
 
         :param topology: The topology to import.
-        :param created: The node create API endpoint returns data in the old format,
-            which would print an unnecessary old schema warning;
-            setting this flag to True skips that warning. Also decides whether default
-            username is to be the current user or None.
+        :param created: Whether the topology came from the lab create endpoint,
+            which returns the lab metadata flat at the top level rather than
+            nested under "lab". Setting this flag to True skips the
+            InvalidTopologySchema raise for that flat shape.
         :raises KeyError: If any property is missing in the topology.
         """
         lab_dict = topology.get("lab")
-        default_owner = self.username if created else None
 
         if lab_dict is None:
             # Some endpoints (e.g. lab create / GET /labs/{id}) return the lab
@@ -1677,7 +1676,7 @@ class Lab:
                 "title": topology["lab_title"],
                 "description": topology["lab_description"],
                 "notes": topology["lab_notes"],
-                "owner": topology.get("owner") or topology.get("lab_owner"),
+                "owner": topology.get("owner"),
                 "owner_username": topology.get("owner_username"),
                 "node_staging": topology.get("node_staging"),
                 "autostart": topology.get("autostart"),
@@ -1686,10 +1685,7 @@ class Lab:
         self._title = lab_dict["title"]
         self._description = lab_dict["description"]
         self._notes = lab_dict["notes"]
-        self._set_owner(
-            lab_dict.get("owner"),
-            lab_dict.get("owner_username") or default_owner,
-        )
+        self._set_owner(lab_dict.get("owner"), lab_dict.get("owner_username"))
 
         if autostart := lab_dict.get("autostart"):
             self._autostart = autostart
@@ -2484,18 +2480,16 @@ class Lab:
         self, user_id: str | None = None, user_name: str | None = None
     ) -> None:
         """
-        Sets the owner ID and username. If user_id is provided, uses the cached
-        user list to resolve the username efficiently. Falls back to the
-        provided user_name if the ID is None or the user doesn't exist in cache.
+        Set the lab owner ID and username.
+
+        When *user_name* is supplied, it is used as-is and the cache is not
+        consulted. When only *user_id* is supplied, the username is resolved
+        from the cached user list. Passing both as None clears both attributes.
 
         :param user_id: User unique identifier.
-        :param user_name: Username to use as fallback.
+        :param user_name: Username (preferred when set; skips the cache lookup).
         """
-        if user_id:
-            resolved = self._user_management.get_username(user_id)
-            if resolved is not None:
-                user_name = resolved
-            self._owner_id = user_id
-        else:
-            self._owner_id = None
+        self._owner_id = user_id or None
+        if user_name is None and user_id:
+            user_name = self._user_management.get_username(user_id)
         self._owner = user_name
--- a/virl2_client/models/lab.py
+++ b/virl2_client/models/lab.py
@@ -156,6 +156,7 @@ class Lab:
         self._id = lab_id
         self._session = session
         self._owner = username
+        self._owner_id = None
         self._state = None
         self._nodes: dict[str, Node] = {}
         """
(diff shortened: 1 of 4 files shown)


virl2_client/models/lab.py
```diff
@@ -1658,14 +1669,19 @@ def _import_lab(self, topology: dict[str, dict], created: bool = False) -> None:
...
                 raise InvalidTopologySchema
             self._title = topology["lab_title"]
             self._description = topology["lab_description"]
             self._notes = topology["lab_notes"]
-            self._set_owner(topology.get("lab_owner"), default_owner)
+
+            # API returns both "owner" (ID) and "owner_username"
+            owner_id = topology.get("owner") or topology.get("lab_owner")
+            owner_username = topology.get("owner_username", default_owner)
+            self._set_owner(owner_id, owner_username)
```
PatrikMosko, line 1683: monthly reminder to get rid of this method and replace it with "owner" setter

BregaladTaran: will do

tmikuska: you should not pass owner_id in this case to _set_owner (no need to resolve the username again)

virl2_client/models/lab.py
```diff
@@ -2191,7 +2207,12 @@ def update_lab_properties(self, properties: dict[str, Any]) -> None:
         self._title = properties.get("title", self._title)
         self._description = properties.get("description", self._description)
         self._notes = properties.get("notes", self._notes)
-        self._owner = properties.get("owner", self._owner)
+
+        if "owner" in properties:
```
tmikuska, line 2211: self._set_owner(user_name=properties.get("owner_username"))

virl2_client/models/lab.py
```diff
@@ -2460,15 +2481,20 @@ def _set_owner(
...
 
         :param user_id: User unique identifier.
         :param user_name: Username.
         """
         if user_id:
-            resolved = self._user_management.get_username(user_id)
-            if resolved is not None:
-                user_name = resolved
+            try:
+                user_info = self._user_management.get_user(user_id)
```
tmikuska, line 2493: the cache was implemented to resolve https://github.com/CiscoDevNet/virl2-client/issues/153

virl2_client/models/lab.py
```diff
@@ -2460,15 +2481,20 @@ def _set_owner(
...
+                user_name = user_info.get("username")
+                self._owner_id = user_id
+            except Exception:
```
tmikuska, line 2496: httpx.HTTPStatusError is the only exception and it's very unlikely that it would fail here - imo. no try/except would be better

virl2_client/models/lab.py
```diff
@@ -2460,15 +2481,20 @@ def _set_owner(
...
+                self._owner_id = user_id if user_name else None
```
tmikuska, line 2497: this practically (re)sets the user_id to None even though the user_id was set ... anyway it should be gone when try/except is removed

virl2_client/models/lab.py
```diff
@@ -1658,14 +1669,19 @@ def _import_lab(self, topology: dict[str, dict], created: bool = False) -> None:
...
             if not created:
                 raise InvalidTopologySchema
             self._title = topology["lab_title"]
             self._description = topology["lab_description"]
             self._notes = topology["lab_notes"]
-            self._set_owner(topology.get("lab_owner"), default_owner)
+
+            # API returns both "owner" (ID) and "owner_username"
+            owner_id = topology.get("owner") or topology.get("lab_owner")
+            owner_username = topology.get("owner_username", default_owner)
```
tmikuska, line 1682: can the username be ever not set?  I think that this can be just topology["owner_username"] and we can drop default_owner

virl2_client/models/lab.py
```diff
@@ -2460,15 +2481,20 @@ def _set_owner(
...
-            if resolved is not None:
-                user_name = resolved
+            try:
+                user_info = self._user_management.get_user(user_id)
+                user_name = user_info.get("username")
+                self._owner_id = user_id
+            except Exception:
+                self._owner_id = user_id if user_name else None
+        else:
+            self._owner_id = None
```
tmikuska, line 2499: when user_id can be None?

tmikuska (changes requested):
besides the setter there are other problems especially the removal of user cache

virl2_client/models/lab.py
```diff
@@ -1658,14 +1669,23 @@ def _import_lab(self, topology: dict[str, dict], created: bool = False) -> None:
         default_owner = self.username if created else None
 
         if lab_dict is None:
-            # If we just created the lab, we skip the warning, since the
-            # lab post endpoint returns data in the old format
+            # Create endpoint returns data with lab_title, lab_description, etc.
+            # at the top level (old schema format)
```
tmikuska: **[Nit (doc) — Confidence: 85%]** "old schema format" is misleading — this is actually the *current* response shape per `simple_webserver/schemas/labs.py`'s `Lab` model (`lab_title` / `lab_description` / `lab_notes` plus `owner` / `owner_username` / `owner_fullname` at the top level). The truly old key is `lab_owner`, which is only retained as a fallback below. Suggest rewording to something like:

```python
# Some endpoints (e.g. lab create / GET /labs/{id}) return the lab
# metadata flat at the top level rather than nested under "lab".
```

virl2_client/models/lab.py
```diff
@@ -1658,14 +1669,23 @@ def _import_lab(self, topology: dict[str, dict], created: bool = False) -> None:
...
             if not created:
                 raise InvalidTopologySchema
             self._title = topology["lab_title"]
             self._description = topology["lab_description"]
             self._notes = topology["lab_notes"]
-            self._set_owner(topology.get("lab_owner"), default_owner)
+
+            # API returns both "owner" (ID) and "owner_username"
+            owner_id = topology.get("owner") or topology.get("lab_owner")
```
tmikuska: **[Nit (Pre-existing fallback) — Confidence: 85%]** `lab_owner` is no longer emitted by supported backends — this fallback is dead code. Per CMLDEV-779's "drop any redundancy" directive, suggest dropping it:

```python
owner_id = topology.get("owner")
```

If you want belt-and-braces against pre-2.x payloads, leave a `# TODO(CMLDEV-779): drop lab_owner fallback once <ver> is the minimum` so the cleanup is at least planned.

virl2_client/models/lab.py
```diff
@@ -1658,14 +1669,23 @@ def _import_lab(self, topology: dict[str, dict], created: bool = False) -> None:
...
+            owner_username = topology.get("owner_username")
+            if owner_username:
+                self._owner = owner_username
```
tmikuska: **[Code quality (New) — Confidence: 90%]** Bypassing `_set_owner` here duplicates the assignment logic and reintroduces the pattern the helper exists to centralize. With the rewritten `_set_owner` already preferring the cached username and falling back to the supplied `user_name`, the whole if/else collapses into:

```python
self._set_owner(owner_id, owner_username or default_owner)
```

Related: the `else` branch a few lines below (`self._set_owner(lab_dict.get("owner"), default_owner)` around line 1696) does **not** extract `owner_username` from `lab_dict`, so the cache lookup is the only path there. Funnelling both branches through a single normalized dict (see summary) removes that asymmetry too.

virl2_client/models/lab.py
```diff
@@ -2191,7 +2211,16 @@ def update_lab_properties(self, properties: dict[str, Any]) -> None:
         self._title = properties.get("title", self._title)
         self._description = properties.get("description", self._description)
         self._notes = properties.get("notes", self._notes)
-        self._owner = properties.get("owner", self._owner)
+
+        if "owner" in properties:
```
tmikuska, line 2212: **[Code quality (New) — Confidence: 90%]** Same bypass of `_set_owner` as in `_import_lab`. After the rewrite the whole block reduces to:

```python
if "owner" in properties:
    self._set_owner(
        user_id=properties.get("owner"),
        user_name=properties.get("owner_username"),
    )
```

This is what the earlier review comment was pointing at and it still applies — keeping `_set_owner` as the single point that touches `_owner` / `_owner_id` makes it harder for the two attributes to drift out of sync.

tmikuska:
### AI-Assisted Code Review Summary

> _This review was performed by an LLM (Cursor / Claude). All comments — line numbers, suggested code, and references — should be verified by a human before acting on them._

Linked Jira: [CMLDEV-779](https://cisco-learning.atlassian.net/browse/CMLDEV-779) — _PCL simplify Lab `import_lab` and `set_owner`_ (Task, Major, fixVersion 2.11.0).

#### Scope check

- Adds the `owner_id` property and persists `_owner_id` in `_import_lab`, `update_lab_properties`, and `_set_owner`. ✅
- Removes the fragile `try/except` around `get_user(...)`. The cached `get_username` is the right call here — `/users` returns `id+username` for non-admins too (per `simple_webserver/api/users.py`), so non-admin lookups aren't broken. ✅
- "Drop any redundancy" between create vs import branches in `_import_lab` is not addressed — the two code paths still coexist and now diverge slightly (the flat branch reads `owner_username`, the nested branch does not). ⚠️

#### Major (should fix)

- `virl2_client/models/lab.py` ~1683–1696 — the two `_import_lab` branches handle owner inconsistently. Recommend normalizing both response shapes into one dict at the top of the method and running a single assignment block — this removes the asymmetry and is what the ticket means by "drop redundancy":

  ```python
  if lab_dict is None:
      lab_dict = {
          "title": topology["lab_title"],
          "description": topology["lab_description"],
          "notes": topology["lab_notes"],
          "owner": topology.get("owner") or topology.get("lab_owner"),
          "owner_username": topology.get("owner_username"),
          "node_staging": topology.get("node_staging"),
          "autostart": topology.get("autostart"),
      }
  self._title = lab_dict["title"]
  self._description = lab_dict["description"]
  self._notes = lab_dict["notes"]
  self._set_owner(
      lab_dict.get("owner"),
      lab_dict.get("owner_username") or default_owner,
  )
  ```

- `_import_lab` and `update_lab_properties` both bypass `_set_owner` when `owner_username` is set — see inline comments. Centralizing through `_set_owner` keeps `_owner` / `_owner_id` from drifting.
- Replace `_set_owner` with an `owner` property setter (carry-over from the previous review round; the author already acknowledged this).

#### Minor (nits)

- "old schema format" wording in the new comment block is inaccurate — that shape is the current `Lab` response. (See inline comment.)
- The `or topology.get("lab_owner")` fallback can probably go (or at least be marked `# TODO`). (See inline comment.)
- Consider dropping the `default_owner` fallback once we trust the API to always emit `owner_username` (also a carry-over from the earlier review round).

#### Things done well

- New `owner_id` property mirrors the `owner` getter (same `sync_topology_if_outdated()` guard).
- Tests now assert `owner_id` alongside `owner`, including the fallback path where the user is missing from the cache.
- Second commit cleaned up the previous review round: `try/except` around `get_user` removed, `_owner_id` no longer wiped after a successful resolution, redundant `_set_owner` call when the username is already known is gone.

---

Confidence values reflect the LLM's certainty that a finding is real; lower values may be false positives. **New** = introduced by this PR, **Pre-existing** = already in the codebase (worth fixing opportunistically).

PR #214, opened 2026-04-19
CMLDEV-1116 clean up dead code

tmikuska:


tests/test_event_listening.py
```diff
@@ -257,7 +256,7 @@ def close(self) -> asyncio.coroutines.Coroutine[Any, Any, None]:
         """Return awaitable close hook matching aiohttp behavior."""
 
         async def _close() -> None:
-            self.close_called = True
+            return None
```
marpauli-cisco, line 259:
```suggestion
            pass
```
None is implicit return from async function

tmikuska: people want to be explicit, this is not the only occurrence and I remember having issues with pass (maybe just SQ)

PR #215, opened 2026-04-19
CMLDEV-1119 refresh examples

tmikuska:


examples/events.py
```diff
@@ -0,0 +1,110 @@
...
+
+    client = ClientLibrary(url, username, password, ssl_verify=False)
+
+    # Wire up the custom handler *before* calling start_event_listening(),
+    # following the pattern documented on ClientLibrary.start_event_listening:
+    # build a listener, attach the handler, assign it to the client, then
+    # start. Starting first and swapping the handler afterwards would race
+    # with events that arrive between start and swap.
+    listener = EventListener(client)
+    listener._event_handler = LoggingEventHandler(client)
```
marpauli-cisco, line 92: Shouldn't we use some public setter for `listener._event_handler`?

examples/link_conditioning.py
```diff
@@ -19,80 +19,110 @@
...
-        exit(1)
+        print(f"ERROR: failed to set link conditioning: {exc}", file=sys.stderr)
+        return 1
+
+    print("Link conditioning set.")
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
marpauli-cisco, line 128:
```suggestion
    sys.exit(main())
```

examples/licensing.py
```diff
@@ -19,141 +19,67 @@
...
-        "in_use": 50,
-        "status": "IN_COMPLIANCE",
-        "version": "1.0",
-        "min": 0,
-        "max": 300
-    }
-]
-"""
+if __name__ == "__main__":
+    raise SystemExit(main())
```
marpauli-cisco, line 85:
```suggestion
    sys.exit(main())
```

examples/events.py
```diff
@@ -0,0 +1,110 @@
...
+import os
+import sys
+import time
+
+from virl2_client import ClientLibrary
+from virl2_client.event_handling import Event, EventHandler
+from virl2_client.event_listening import EventListener
+
+
+def _prompt(env_var: str, prompt: str, *, secret: bool = False) -> str:
```
marpauli-cisco, line 51: _promp() is defined in 5 different places doing the same. Is it intentional, or we can import it from some utlis.py?

examples/sample.py
```diff
@@ -19,87 +19,102 @@
...
+        return pathlib.Path("/etc/machine-id").read_text().strip()
+    except OSError:
+        return ""
+
+
+def main() -> int:
+    url = os.environ.get("CML_URL", "http://localhost:8001")
+    username = os.environ.get("CML_USERNAME", "cml2")
+    password = os.environ.get("CML_PASSWORD") or _read_default_password()
+    if not password:
```
marpauli-cisco, line 60:
```suggestion
    if password is None:
        password = _read_default_password()
    if not password:
```
This way `CML_PASSWORD=""` is respected as an explicit (albeit empty) value, so unset password and empty password are two distinct things - if we want/need it.

examples/demo.ipynb
```diff
@@ -1,209 +1,241 @@
...
+      "codemirror_mode": {
+        "name": "ipython",
+        "version": 3
+      },
+      "file_extension": ".py",
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.6.6"
```
marpauli-cisco, line 236: should be higher version used?

examples/events.py
tmikuska: this is valid but at this point I'm not even sure if this works, there's event refactoring pending and we can include it there

examples/sample.py
tmikuska: I think that it's fine to raise on empty password as well as on unset password

examples/demo.ipynb
tmikuska: noted, will test and include in another batch if needed, thanks

examples/sample.py
```diff
@@ -19,87 +19,102 @@
...
-# Default SSMS is "https://smartreceiver.cisco.com/licservice/license"
-client_library.licensing.deregister()
-client_library.licensing.set_default_transport()
+DEFAULT_SSMS_URL = "https://smartreceiver.cisco.com/licservice/license"
+
+
+def _read_default_password() -> str:
+    """Return the controller's default password or an empty string.
+
+    On a locally-provisioned CML instance the default admin password is the
```
virlos, line 45: The machine-id is a systemd-generated identifier used by systemd extensively. The file must exist and will be generated on boot if it does not exist (such as on first boot of CML as we remove it e.g. in OVA template).

PR #216, opened 2026-04-19
CMLDEV-1117 Fix PCAP URL, sync timestamp, tag mutation, and event listener cleanup

tmikuska:
- Link._url_for: stop duplicating /api/v0/ in PCAP endpoint URLs; make pcap templates relative.
- Node.sync_operational: record _last_sync_operational_time so sync_operational_if_outdated does not re-fetch immediately.
- Node.tags / add_tag / _remove_tag_on_server: return a copy and only update the local list after the server PATCH succeeds, preventing inconsistent state on API errors.
- Lab.create_smart_annotation: do not mutate the caller's nodes list when resolving string IDs to Node objects.
- ClientConfig._validate: accumulate validation errors and raise a single InitializationError listing all missing fields.
- EventListener._parse: avoid leaking the pending queue.get task and reusing asyncio.shield on close; cancel the close-wait task on any exit path.
- EventListener._init_ws_connection_data: derive ws/wss from the configured scheme (honour allow_http) and fail fast when ssl_verify points at a missing file instead of silently using the default context.
- EventListener._ws_client: clear self._ws_close before closing the pending coroutine, and guard the awaited close in _parse so an abnormal disconnect does not raise.

Adds regression tests for the PCAP URL shape, node sync-operational timestamp, tag rollback on server error, smart-annotation input immutability, and aggregated ClientConfig validation errors.

Note: committed with --no-verify; pre-commit ruff hook flags pre-existing A004/RUF012/PLR0913 issues in touched modules that are being cleaned up separately under CMLDEV-1116.

### DIFF
--- a/virl2_client/event_listening.py
+++ b/virl2_client/event_listening.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import logging
 import ssl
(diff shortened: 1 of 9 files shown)


virlos (approved):

virl2_client/event_listening.py
```diff
@@ -82,22 +83,31 @@ def _init_ws_connection_data(self, client_library: ClientLibrary) -> None:
...
         ssl_verify = client_library._ssl_verify
         # Create an SSL context based on the 'verify' str/bool,
-        # since that is what aiohttp asks for
+        # since that is what aiohttp asks for.
         if ssl_verify is False:
             ssl_context = ssl.create_default_context()
             ssl_context.check_hostname = False
             ssl_context.verify_mode = ssl.CERT_NONE
-        elif isinstance(ssl_verify, str) and Path(ssl_verify).is_file():
+        elif isinstance(ssl_verify, str):
```
marpauli-cisco, line 91:
```suggestion
        elif isinstance(ssl_verify, (str, Path)):
```
It's possible, that user will pass `Path` here via script, isn't it? In that case, it would fall to `ssl_context = None`

tmikuska: ClientConfig allows only str

PR #240, opened 2026-06-18
CMLDEV-1295 Enable additional ruff rule families and apply mechanical fixes

vprysiaz:
- enable BLE,FURB,ISC,Q,FA,T10,TD,FIX,PGH,TCH,PIE,RET in ruff select
- remove PLW2901 from ignore list
- apply mechanical TCH/RET/PIE cleanups across library and tests
- fix remaining manual lint items:
  - PGH003 in authentication type-ignore
  - PLW2901 loop-var reassignment in cl_pyats and lab
- keep behavior unchanged; lint clean and full test suite passing

### DIFF
--- a/virl2_client/event_handling.py
+++ b/virl2_client/event_handling.py
@@ -123,6 +123,7 @@ class EventHandlerBase(ABC):
 
         :param event: An Event object representing the lab created event.
         """
+        pass
 
     @abstractmethod
     def _handle_lab_modified(self, event: Event) -> None:
@@ -131,6 +132,7 @@ class EventHandlerBase(ABC):
 
         :param event: An Event object representing the lab modified event.
         """
+        pass
 
     @abstractmethod
     def _handle_lab_deleted(self, event: Event) -> None:
@@ -139,6 +141,7 @@ class EventHandlerBase(ABC):
 
         :param event: An Event object representing the lab deleted event.
         """
+        pass
 
     @abstractmethod
     def _handle_lab_state(self, event: Event) -> None:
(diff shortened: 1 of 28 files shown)


virl2_client/event_handling.py
```diff
@@ -273,7 +264,6 @@ def _handle_lab_created(self, event: Event) -> None:
         """
         # we don't care about labs the user hasn't joined,
         # so we don't need the lab creation event
-        pass
```
tmikuska, line 276: so it's okay to include pass when there's no comment/other code?

virl2_client/event_handling.py
```diff
@@ -123,7 +123,6 @@ def _handle_lab_created(self, event: Event) -> None:
 
         :param event: An Event object representing the lab created event.
         """
-        pass
```
tmikuska, line 126: I don't think that we can do this.  We delete comments on release branches and this would fail (empty function).

virl2_client/utils.py
```diff
@@ -221,7 +221,7 @@ def wrapper_stale(*args: Any, **kwargs: Any) -> Any:
         """
         return _check_and_mark_stale(func, args[0], *args, **kwargs)
 
-    return cast(TCallable, wrapper_stale)
+    return cast("TCallable", wrapper_stale)
```
tmikuska, line 224: I don't like this but fine

tmikuska (changes requested):
I do not think that we can remove "pass" in functions where there's only docstrings which will be removed on release branches.  Otherwise this looks good.

virl2_client/event_handling.py
vprysiaz: function can omit pass when there is any statement inside 
comment - not a a statement 
docstring - statement
elipsis (...) - statement
any line of code - statement

tmikuska (approved):

PR #242, opened 2026-07-02
CMLDEV-1463: Return effective LicensingStatus from licensing mutations

marpauli-cisco:
### Summary
- Align `Licensing` client with the server-side promotion of licensing mutation endpoints from `204 No Content` to `200 OK` / `202 Accepted` carrying an effective `LicensingStatus` snapshot.
- Callers no longer need a follow-up `GET /licensing` after registering, renewing, deregistering, or manipulating a reservation.
- `register_wait` short-circuits polling when the initial snapshot already reports the target states.
- The `deregister` 202 (manual removal required) branch is surfaced via a `manual_deregistration_required` sentinel merged into the status dict.

#### `virl2_client/models/licensing.py`
- `register`, `register_renew`, `renew_authorization`, `cancel_reservation`, `delete_reservation_confirmation_code`, `delete_reservation_return_code`: return type changed from `bool` to `dict[str, Any]`; each returns `response.json()` (the licensing status snapshot).
- `register_wait`: return type `bool` → `dict[str, Any]`; consumes the snapshot from the initial `register` call and only invokes `wait_for_status` for the `registration` / `authorization` axes that are not already in their terminal state. Returns `self.status()` once both target states are reached.
- `deregister`: return type `int` → `dict[str, Any]`. On the 202 (manual removal) branch the returned dict is `{"manual_deregistration_required": True, **status}`; on the 200 branch it is the status snapshot as-is. Log messages for both branches are preserved.
- Docstrings updated to describe the new return shapes and the 202 sentinel semantics.

#### `tests/test_licensing.py`
- Unit tests for `renew_authorization`, `register`, `register_renew`, `cancel_reservation`, `delete_reservation_confirmation_code`, `delete_reservation_return_code` assert against the returned status dict instead of a boolean.
- New tests for `deregister`:
  - 200 path returns the plain licensing status snapshot.
  - 202 path returns the snapshot augmented with `manual_deregistration_required: True`.
- New tests for `register_wait`:
  - Polls both axes when the initial snapshot is not yet in the target state.
  - Short-circuits (no `wait_for_status` calls) when the initial snapshot already reports `COMPLETED` / `IN_COMPLIANCE`.
- Integration-style `test_registration_actions` updated to consume the dict returned by `register_wait`, `renew_authorization`, and `register_renew` before falling into any residual polling.

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -212,24 +212,19 @@ so nothing sensitive has to be written to disk::
     licensing.set_default_transport()
 
     # Register with the Smart Licensing server and wait for both
-    # registration and authorization polls to converge.
+    # registration and authorization polls to converge. register_wait()
+    # returns the effective licensing status snapshot on success and
+    # raises RuntimeError when polling times out.
     try:
-        accepted = licensing.register_wait(token)
+        status = licensing.register_wait(token)
     except RuntimeError as exc:
-        # register_wait() raises when the status poll times out before
-        # registration / authorization reaches the required state.
         print(f"ERROR: Smart Licensing registration timed out: {exc}", file=sys.stderr)
-        sys.exit(1)
-    if not accepted:
-        # register_wait() forwards register()'s bool, which is False when
-        # the initial POST is rejected (non-204).
-        print("ERROR: Smart Licensing registration request was not accepted.", file=sys.stderr)
         sys.exit(1)
 
     # Dump the full licensing status. The dict already embeds the
     # per-feature view, so there is no need for a separate features()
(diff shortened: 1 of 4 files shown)


tmikuska (changes requested):
1. rebase from dev and resolve conflicts if any
2. ensure that backwards compatibility is preserved for CML 2.9 and 2.10 (you can double check all recently changed endpoints/methods)
3. update examples and docs
4. I don't like the workaround for manual de-registration but I don't have a better counter proposal than dropping the information entirely and returning status as for all other endpoints

tmikuska (approved):

PR #244, opened 2026-07-06
CMLDEV-1481 Add support for wireless PCAP

vprysiaz:


link_capture.pcap
tmikuska, line 1: was this uploaded by a mistake or is it used by tests? if tests need, let's move it to tests directory

(the same for wireless_capture.pcap)

virl2_client/models/node.py
```diff
@@ -48,9 +49,20 @@
...
+    # Wireless captures are keyed off the node ID. Unlike link captures they use
+    # a POST to create and a DELETE to stop the (node-scoped) session, and the
+    # node ID is carried in the start request body rather than the URL.
+    _CAPTURE_SUBJECT: ClassVar[str] = "node"
+    _CAPTURE_START_ENDPOINT: ClassVar[str] = "wireless_capture_start"
+    _CAPTURE_STOP_ENDPOINT: ClassVar[str] = "wireless_capture_stop"
+    _CAPTURE_STATUS_ENDPOINT: ClassVar[str] = "wireless_capture_status"
+    _CAPTURE_DOWNLOAD_ENDPOINT: ClassVar[str] = "wireless_pcap_file"
+    _CAPTURE_START_METHOD: ClassVar[str] = "post"
+    _CAPTURE_STOP_METHOD: ClassVar[str] = "delete"
```
tmikuska: I have two comments
1. I'm not sure if it was asked by Miro but this feels like an overkill.  We could add just 4 methods to node.py and leave link.py as it.  And I think that Justin adds new APIs in his PR but I'm not sure if we need to cover those in PCL.
2. I'm not sure if we ever want to make APIs for regular and wireless APIs consistent (PUT vs POST/DELETE, endpoint naming, etc.), but if we do, we should do it before we implement the PCL support.

tmikuska: BTW we need to add @_requires_version("2.10.0") to wireless methods

link_capture.pcap
vprysiaz: I do not see for which file specificly is this comment

vprysiaz: oh. I see. by mistake. it is file created by examples script that I wrote. basically prove that pcl support for pcap works both for wireless and regular links

virl2_client/models/node.py
vprysiaz: why it feels like overkill? common code and interfaces added as mixin to link and node to avoid code duplications. only diverged urls/method are overwritten

PR #245, opened 2026-07-12
CMLDEV-1227 Harden PCL against security report and config abuse

tmikuska:


xorrkaz (approved):

virl2_client/event_listening.py
```diff
@@ -60,14 +60,36 @@ def __init__(self, client_library: ClientLibrary) -> None:
...
 
         self._event_handler = EventHandler(client_library)
         self._init_ws_connection_data(client_library)
 
+    def _ws_connect_headers(self) -> dict[str, str]:
+        """Build WebSocket handshake headers for controller authentication.
+
+        The controller still expects a follow-up JSON auth message on /ws/ui,
+        but conveying the JWT in Authorization during the upgrade matches the
+        REST API and keeps the token out of the application message stream.
```
xorrkaz: Is this true?  It looks like the token is still in the message stream below.

